PR #5, opened 2024-09-13
Kuberay solution with head issues fixes; under modules/kuberay/main.t…

idanb1:
Kuberay solution with head issues fixes; under modules/kuberay/main.tf added a temporary csi-driver based on csi-hostpath by OlegF (requirement to deploy ray-redis pods); tested on nebius provider version 0.3.19; Kuberay autoscaling added and tested + README.MD updated with autoscaling test example; uncommenting the Cloud environment from .tfvars file + removing the work 'OR' with relevant comments.


k8s-training/terraform.tfvars
```diff
@@ -1,11 +1,11 @@
 # Cloud environment and network
-# parent_id      = "" # The project-id in this context
-# subnet_id      = "" # Use the command "nebius vpc v1alpha1 network list" to see the subnet id
-# ssh_user_name  = "" # Username you want to use to connect to the nodes
-# ssh_public_key = {
-# key  = "put your public ssh key here" OR
-# path = "put path to ssh key here"
-# }
+parent_id      = "" # The project-id in this context
```
d3vil-st, line 2: Empty strings as values will broke tests and could lead con undefined behavior in real usage.

modules/kuberay/main.tf
```diff
@@ -0,0 +1,61 @@
...
+  
+  # Flatten the list of lists into a single list of YAML documents
+  flat_yaml_contents = flatten(local.yaml_contents)
+  
+  # Remove any empty documents
+  cleaned_yaml_contents = [for doc in local.flat_yaml_contents : doc if length(regexall("^\\s*$", doc)) == 0]
+}
+
+# Running 'kubectl apply -f ./csi-filestore/deploy' to set csi-driver & dependencies
+resource "kubectl_manifest" "csi_filestore_v2" {
```
d3vil-st, line 17: You could use official '[manifest](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/manifest)' resource from hashicorp here, to prevent pulling new dependency.

modules/kuberay/main.tf
```diff
@@ -0,0 +1,61 @@
...
+
+resource "helm_release" "kuberay-operator" {
+  name             = var.kuberay_name
+  repository       = var.kuberay_repository_path
+  chart            = var.kuberay_chart_name
+  namespace        = var.kuberay_namespace
+  create_namespace = var.kuberay_create_namespace
+  version          = "1.1.0"
+  values           = [
+    "${file("../modules/kuberay/helm/ray-values.yaml")}"
```
d3vil-st, line 34: This will not work for standalone module usage, `"${path.module}/files/ray-values.yaml"` variable could be used here.

modules/kuberay/variables.tf
```diff
@@ -0,0 +1,67 @@
+variable "folder_id" {
+  description = "The ID of the folder that the Bastion should be deployed to."
+  type        = string
+  default     = null
```
d3vil-st, line 4: If `null` value for variable breaks module, better avoid `default` property for it.

modules/kuberay/main.tf
idanb1: Since this solution wasn't that stable, I took mk8s helm release solution and integrated it in our terraform k8s-training.

It will be part of my next commit + other fixes

modules/cloud-init/k8s-cloud-init.tftpl
```diff
@@ -2,9 +2,9 @@
 bootcmd:
 %{ endif }
 %{ if enable_filestore != "false" }
- - sudo mkdir -p /mnt/filestore
- - sudo mount -t virtiofs filestore-device /mnt/filestore
- - echo filestore-device /mnt/filestore "virtiofs" "defaults" "0" "2" | sudo tee -a /etc/fstab
+- sudo mkdir -p /mnt/data
```
d3vil-st, line 5: Could you please check indent here, I suppose this could break yaml syntax.

k8s-training/provider.tf
```diff
@@ -2,6 +2,7 @@ terraform {
   required_providers {
     nebius = {
       source = "terraform-provider-nebius.storage.ai.nebius.cloud/nebius/nebius"
+      version = "0.3.19"
```
d3vil-st, line 5: Are we locking provider version now?

modules/csi-mounted-fs-path/versions.tf
d3vil-st, line 1: I think we don't need this file.

modules/kuberay/versions.tf
d3vil-st, line 1: We could avoid version locking here.

k8s-training/provider.tf
idanb1: I can remove it, but last time I tried, there were still issues with .22

modules/kuberay/versions.tf
idanb1: For both nebius and helm?

d3vil-st (approved):

malibora (approved):

PR #19, opened 2024-09-30
Updating Solutions to use migrated NVIDIA Operators;

elijah-k-nebius:


.github/workflows/terraform.yml
```diff
@@ -24,11 +24,11 @@ jobs:
       max-parallel: 2
       matrix:
         solution:
-          - name: compute-testing
```
d3vil-st: I think we can remove this one, since it's not deploying anything.

PR #35, opened 2024-10-09
Transition to applications from helm for applications products;

elijah-k-nebius:


d3vil-st (approved):

modules/gpu-operator/helm.tf
d3vil-st, line 1: Should we rename this file to `main.tf`?

modules/network-operator/helm.tf
d3vil-st, line 1: Should we rename this file to `main.tf`?

PR #41, opened 2024-10-16
Make use of LoadBalancer service type

dstaroff:


rdjjke (approved):

soperator/modules/login/templates/login.sh.tftpl
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+set -e
+
+usage() { echo "usage: $${0} [-u <ssh_user>] -k <path_to_ssh_key> [-h]" >&2; exit 1; }
```
asteny, line 5: Where is an address here?

asteny: And port

asteny (changes requested):

soperator/modules/login/templates/login.sh.tftpl
dstaroff: Address and port are being set via templating.
There's a little difference in varables - bash variables are escaped with additional dollar sign `$${}`, while TF template variables are regular `${}`

asteny (approved):

PR #53, opened 2024-10-24
Soperator recipe versioning

dstaroff:


rdjjke (approved):

asteny:
I your variant we should think remember to modify manually dev version with hash, then return it back. Is that we want?

asteny (approved):

PR #64, opened 2024-10-28
Feature/nfs module

jadnov:
nfs module, standalone and integrated in slurm template


slurm/versions.tf
```diff
@@ -7,5 +7,5 @@ terraform {
 }
 
 provider "nebius" {
-  domain = "api.eu.nebius.cloud:443"
+  domain = "api.eu-north1.nebius.cloud:443"
```
asteny, line 10: ?

slurm/files/ansible_role.zip
d3vil-st, line 1: I would suggest to remove that file and add into .gitignore

slurm/versions.tf
jadnov: mistake, fixing

slurm/files/ansible_role.zip
jadnov: Ilya, you forget this is actually an ansible file transferred to VM, without it will not work

jadnov (commented):
ok

PR #79, opened 2024-11-13
NOTASK: increse deafult resources vmagent

Uburro:


soperator/modules/monitoring/variables.tf
```diff
@@ -54,8 +54,8 @@ variable "resources_vm_agent" {
     cpu    = string
   })
   default = {
-    memory = "384Mi"
-    cpu    = "250m"
+    memory = "1Gi"
```
asteny, line 57: Should we up default resources in Soperator helm?

Uburro: we do not have vmagent on soperator helm

asteny (approved):

dstaroff (approved):

PR #81, opened 2024-11-13
Soperator deployment fixes

dstaroff:
- Filter service accounts with jq, as filtering via API is not implemented
- Fix for division worker nodeset into node groups
- Remove setting K8s version for node groups
- Do not access accounting resources when it's disabled
- Remove registry from version sync
- Add support for new platforms


asteny (approved):

alex000kim:
I can confirm that this fixes that issue.
A request though:
Can we rename size in nodeset configs to `worker_count` (or something with `count`) to make it explicit.
Personally, I'd prefer to keep `size` for storage/RAM related configs.
Or `nodes_count` to be more consistent with other solutions:
https://github.com/search?q=repo%3Anebius%2Fnebius-solution-library%20nodes_count&type=code

rdjjke (approved):

rdjjke (approved):

PR #83, opened 2024-11-14
MSP-3313: Add NSYS profiling in GPT3 mlperf implementation

rdjjke:


soperator/mlperf/gpt3-impl-4.0-nvidia/Dockerfile
```diff
@@ -121,6 +121,21 @@ RUN rm -rf /opt/hpcx/nccl_rdma_sharp_plugin && \
 RUN pip install huggingface_hub==0.23.2
 RUN pip install -v "transformers<=4.40.2"
 
+## Reinstall NCCL to the latest version
```
asteny, line 124: Why that commented?

rdjjke: Because it's not used by default. Just saved it because I may need it in the future

PR #98, opened 2024-11-26
Fix ubuntu username with ssh_user_name variable

sazriel26:
At least, default user name could change to something like `nebius` 🤗 

Kindly review if it can suits (terraform test of wireguard is more complicated since involving ssh private key detection, but why not - tested with below configuration)

```
...
ssh_public_key = {
# key  = "put your public ssh key here" OR
  path = "~/.ssh/id_nebius.pub"
}
...
```


malibora:
slurm solution is deprecated. thanks anyway @sazriel26 !

PR #116, opened 2024-12-16
add protected_secret

Uburro:


soperator/modules/slurm/variables.tf
```diff
@@ -248,6 +248,12 @@ variable "accounting_enabled" {
   default     = false
 }
 
+variable "protected_secret" {
```
dstaroff, line 251: Let's rename it to something like `use_protected_secret` or `protect_secret`. Current name is misleading as it looks like some secret is expected rather than flag to protect it or not

soperator/modules/slurm/variables.tf
```diff
@@ -248,6 +248,12 @@ variable "accounting_enabled" {
...
+  description = "Protected secret for MariaDB."
```
asteny, line 252: Let's put here a small description of what that means

asteny (approved):

dstaroff (changes requested):

asteny (approved):

soperator/modules/slurm/variables.tf
```diff
@@ -248,6 +248,12 @@ variable "accounting_enabled" {
   default     = false
 }
 
+variable "use_protected_secret" {
+  description = "If true, protected secret MariaDB will be deleted after the MariaDB CR is deleted."
```
dstaroff, line 252: I don't understand why is it called `protected secret` if it is being deleted with CR

PR #125, opened 2024-12-20
Precise allocatable ephemeral storage calculation

dstaroff:


soperator/installations/example/main.tf
```diff
@@ -206,32 +206,54 @@ module "slurm" {
...
-        cpu_cores                   = local.resources.workers[i].cpu_cores
-        memory_gibibytes            = local.resources.workers[i].memory_gibibytes
-        ephemeral_storage_gibibytes = ceil(worker.boot_disk.size_gibibytes - 2 * module.resources.k8s_ephemeral_storage_reserve.gibibytes)
-        gpus                        = local.resources.workers[i].gpus
+        cpu_cores        = local.resources.workers[i].cpu_cores
+        memory_gibibytes = local.resources.workers[i].memory_gibibytes
+        ephemeral_storage_gibibytes = (
+          module.k8s.gpu_involved
+          ? floor(
+            module.resources.k8s_ephemeral_storage_coefficient * worker.boot_disk.size_gibibytes
```
rdjjke, line 231: There are pods other than slurmd on worker nodes (i.e. munge), which also request some resources. Is this counted here?

soperator/modules/available_resources/main.tf
```diff
@@ -251,5 +251,5 @@ locals {
 }
 
 data "units_data_size" "k8s_ephemeral_storage_reserve" {
-  gibibytes = 64
+  gibibytes = 48
```
rdjjke, line 254: I suggest to set this to 64, just in case

soperator/installations/example/main.tf
dstaroff: Yup. This is a total amount of ephemeral storage available for a worker pod. 
- Here we have a pre-defined list of resources taken by sidecars/containers/pods: https://github.com/nebius/nebius-solution-library/blob/a86fc72a7e515476dce423dac50cb257a34fe0a9/soperator/modules/slurm/locals.tf#L60
- Here we set available storage size without such entities: https://github.com/nebius/nebius-solution-library/blob/5ff42b20670d1724a3516028eb13b29b20b759ab/soperator/modules/slurm/main.tf#L211

soperator/modules/available_resources/main.tf
dstaroff: Done

rdjjke (approved):

PR #127, opened 2024-12-23
Unification of test recipes

dstaroff:
This involves unification of:
- Test data and scripts delivery process
- MLCommons SD & GPT3 benchmark recipes
- Quickhecks
- Help messages

This is now done in similar manner with little to zero needed user-facing configuration.
It also contains resurrected docs on test running process and some bug fixes and enhancement on TF recipe configuration.


soperator/test/common/enroot.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+ENROOT_CONFIG_DIR='/root/.config/enroot'
```
asteny, line 3: Should we run test from user root? I implemented running enroot from different users

rdjjke (approved):

soperator/test/common/enroot.sh
dstaroff: Enroot configuration could be removed at all later, as it's not required to configure it

PR #133, opened 2024-12-23
[ADD] Set GPU Operator driver version to 550.127.08

gabdrakhmano:
"For Soperator, it’s better to set the default to this new value 550.127.08 and periodically update it manually, because for some version upgrades we also need to upgrade the CUDA version." (c) Mikhail Mokrushin


dstaroff (approved):

dstaroff:
Let's wait for LLM team to test this version first.

dstaroff:
I tried installing this version of driver while tested new installation, and it didn't work out of the box. Some research and testing are needed before merging it.

PR #140, opened 2025-01-03
Separate sshd configs for login and worker nodes

asteny:


soperator/modules/slurm/variables.tf
```diff
@@ -11,7 +11,7 @@ variable "operator_version" {
 variable "operator_stable" {
   description = "Whether to use stable version of the Soperator."
   type        = bool
-  default     = true
+  default     = false
```
dstaroff, line 14: Why do we want to use unstable version as default?

dstaroff (changes requested):
The rest LGTM

dstaroff (approved):

PR #150, opened 2025-01-09
Bump 1.17.0

Uburro:


soperator/VERSION
rdjjke, line 1: Have you also set SUBVERSION to 1?

rdjjke (approved):

PR #155, opened 2025-01-13
[FIX] Get subnet's CIDR from status rather than from spec [MSP-3950]

dstaroff:


Uburro (approved):

soperator/installations/example/main.tf
```diff
@@ -72,18 +72,24 @@ module "filestore" {
...
+  source = "../../../modules/nfs-server"
+
+  parent_id = data.nebius_iam_v1_project.this.id
+  subnet_id = data.nebius_vpc_v1_subnet.this.id
+
+  platform      = var.nfs.resource.platform
+  preset        = var.nfs.resource.preset
+  instance_name = "${local.k8s_cluster_name}-nfs-server"
+
+  nfs_ip_range = data.nebius_vpc_v1_subnet.this.status.ipv4_private_cidrs[0]
```
dstaroff, line 86: https://docs.nebius.com/terraform-provider/reference/data-sources/vpc_v1_subnet#nested-schema-for7

PR #156, opened 2025-01-13
Feature/node group sa

negrovalar:
Please review and run automated tests.
This commit adds a new variable tenant_id that is mandatory to extract group id (parent for groups is a tenant). 
tenant_id      = "" # The tenant id


dstaroff (changes requested):
Hi @negrovalar. `soperator` module has dedicated release process. Please, create another PR for `soperator` and set `release/soperator` as the target branch. Also, run `terraform fmt -recursive` on changed directories

malibora:
@negrovalar  this one should be removed right?

PR #159, opened 2025-01-16
Feature/simple solutions part 1

rene-tech:


malibora (commented):
@rene-tech  the folder is called simple solutions infering there will be more than one, but then in root you have one solution.
Pls organize it in sub folder called something like compute-filestore or something so that we can add more

malibora (approved):

PR #183, opened 2025-02-03
Feature/node group sa main part

negrovalar:
Following up on the previous PR (https://github.com/nebius/nebius-solution-library/pull/156
) I split it into two parts - soperator module and main part.
This is the main part. Please merge it after the soperator part - https://github.com/nebius/nebius-solution-library/pull/182. 
And please run all available auto-tests.


negrovalar:
Correct, no dependency on soperator module. You can merge it at your convenience.

malibora:
@negrovalar   pls run "terraform fmt -recursive" in hte solution so that the tests can pass

negrovalar:
@malibora done, please check

malibora:
@negrovalar  it still fails fmt check in modules folder: 
image

malibora:
@negrovalar  the tests are up anr running again... we have new failed test now: 

image

negrovalar:
@malibora sorry my bad - I haven't done the proper var cleaning after PR split. Please try again

malibora:
@negrovalar we got a new error: 
image


I think the reason is that tests run k8s-training and k8s-inference in parallel so that's why service account already exists.

Additional, the script needs to be indenpotent  even with its self, can you pls add a random sufix so that this works? You can use the one from locals, as we do for other resources: 
https://github.com/nebius/nebius-solution-library/blob/8dcf285514eb0a1f6522f7e31de1123bf63dce94/k8s-training/locals.tf#L2

PR #200, opened 2025-02-20
soperator: rework backups deployment

webconn:
Now backups are deployed on a turn-key basis, including service account, S3 bucket and schedules.


soperator/modules/backups/main.tf
```diff
@@ -0,0 +1,145 @@
...
+              "  annotations:",
+              "    slurm.nebius.ai/service-account: ${self.triggers_replace.service_account_id}",
+              "data:",
+              "  aws-access-key-id: $(nebius iam access-key get-by-id --id $AKID | yq .status.aws_access_key_id | tr -d '\n' | base64)",
+              "  aws-access-secret-key: $(nebius iam access-key get-secret-once --id $AKID | yq .secret | tr -d '\n' | base64)",
+              "  backup-password: $(echo -n ${var.backups_password} | base64)",
+              "\" ;",
+            ]
+          ),
+        " } | kubectl apply --server-side --context ${var.k8s_cluster_context} -f -"
```
Uburro, line 115: let's deploy it namespace other than default

Uburro (dismissed):

soperator/installations/example/terraform.tfvars
rdjjke, line 1: If backups are disabled by default, most deployments won't have them.
Can we implement the following logic?
- `backups_enabled = "default"` - enable backups when at least one `jail_submount` is created
- `backups_enabled = "true"` - force enable backups
- `backups_enabled = "false"` - force disable backups

soperator/installations/example/terraform.tfvars
```diff
@@ -366,9 +368,33 @@ backups_enabled = false
 # ---
 backups_password = "password"
 
-# endregion Backups
+# Cron schedule for backup task.
+# See https://docs.k8up.io/k8up/references/schedule-specification.html for more info.
+# ---
+backups_schedule = "@hourly-random"
```
rdjjke, line 374: It's better to have a good default here too.
What about the following logic?
1. Create a separate global variable `timezone`.
2. Make both `backups_schedule` and `backups_prune_schedule` accept the following values:
    - `"auto"` - create backups every night (e.g. 4 AM) in the specified timezone
    - `"manual:@hourly-random"` or `"manual:52 4 26 * *"` to set a specific schedule

soperator/installations/example/terraform.tfvars
rdjjke: Another option is to create two bool variables instead of adding a string enum:
- `backups_enabled_auto = true` - don't respect `backups_enabled` and set it automatically
- `backups_enabled = true` - when `backups_enabled_auto = false`, this value will enable or disable backups

soperator/installations/example/terraform.tfvars
rdjjke: The same applies here - https://github.com/nebius/nebius-solution-library/pull/200/files#r1966529949

Instead of creating one enum variable, we can use two: `backups_schedule_auto` and `backups_schedule`.

soperator/installations/example/terraform.tfvars
webconn: isn't it an over-complication of things? isn't jail a necessary part of the deployment? if so, we could just enable backups by default

webconn: We decided to enable backups for jails less than 12TB in size by default, fixed it

PR #206, opened 2025-02-24
add nodeconfigurator and soperator checks to slurm module

itechdima:


soperator/modules/slurm/templates/helm_values/node_configurator.yaml.tftpl
```diff
@@ -0,0 +1,18 @@
...
+  logFormat: json
+  image:
+    repository: "${rebooter.image.repository}"
+    tag: "${rebooter.image.tag}"
+    resources:
+      limits:
+        memory: ${rebooter.resources.requests.memory}Gi
+      requests:
+        memory: ${rebooter.resources.requests.memory}Gi
+        cpu: ${rebooter.resources.requests.cpu * 1000}m
```
Uburro, line 16: I do not like this one. What should I do if I want to use 200m cpu and 300Mi memeory?

itechdima: then you set it as decimal (like here):
https://github.com/nebius/nebius-solution-library/blob/main/soperator/modules/slurm/locals.tf#L76

I made it to have unified approach specifying resources for the services

Uburro (approved):

PR #229, opened 2025-03-18
Remove k8s-inference solution

afonsonebius:
As per internal NEBIUS request at https://nebius.atlassian.net/browse/ARCHVTEAMS-663:

"the k8s-training and k8s-inference solutions are too similar to be a different solution. Even more, its misleading because we have clients doing inference on k8s-training solution.

So: remove k8s-inference solution"


aaroniscode (approved):
Approved to run tests

PR #236, opened 2025-03-20
Ainfante nebius remove k8s inference

afonsonebius:


afonsonebius (commented):
Removed references to `k8s-inference` since it was deleted

PR #238, opened 2025-03-21
508: Added values for sconfigcontroller helm

andrei-pokhila:
Added SConfigController values to the helm template

Making changes with the main PR into soperator: https://github.com/nebius/soperator/pull/543

### DIFF
--- a/soperator/modules/slurm/variables.tf
+++ b/soperator/modules/slurm/variables.tf
@@ -366,3 +366,39 @@ variable "enable_soperator_checks" {
 }
 
 # endregion SoperatorChecks
+
+# region SConfigController
+
+variable "sconfigcontroller" {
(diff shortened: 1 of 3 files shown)


soperator/modules/slurm/variables.tf
```diff
@@ -366,3 +366,39 @@ variable "enable_soperator_checks" {
...
+      image_pull_policy = string
+      resources = object({
+        cpu               = number
+        memory            = number
+        ephemeral_storage = number
+      })
+    })
+  })
+  default = {
+    slurm_api_server = "http://soperator-rest-svc.soperator:6820"
```
Uburro, line 390: why we have to set it here and is not slurm_api_server aotogenerated in soperator side?

itechdima (approved):

PR #248, opened 2025-03-26
Fix breaking tests by running `terraform fmt` on modules

aaroniscode:


jadnov (approved):
TF format changes only, approving

PR #256, opened 2025-03-27
fix soperator hello.sh quickcheck

aaroniscode:
Test fails because in `sbatch` the default limit of tasks is 1 and srun use using 2.

Also removing allocation of cpu/mem


aaroniscode:
This was resolved in a different PR

PR #265, opened 2025-03-31
Don't call scontrol from within prolog/epilog scripts to drain the node

dstaroff:
There are few sections in [Slurm doc](https://slurm.schedmd.com/prolog_epilog.html):

> Prolog and Epilog scripts should be designed to be as short as possible and should not call Slurm commands (e.g. squeue, scontrol, sacctmgr, etc). Long running scripts can cause scheduling problems when jobs take a long time to start or finish. Slurm commands in these scripts can potentially lead to performance issues and should not be used.

And we had such issues with Prolog scripts being hung on un-performant nodes:
```
slurmstepd: error: Prolog hung on node worker-X
```

> If the Epilog fails (returns a non-zero exit code), this will result in the node being set to a DRAIN state. If the EpilogSlurmctld fails (returns a non-zero exit code), this will only be logged. If the Prolog fails (returns a non-zero exit code), this will result in the node being set to a DRAIN state and the job requeued.

So, exiting with non-zero code should work the same as with draining the node with `scontrol update`, but should not cause performance issues and hanging scripts.

### DIFF
--- a/soperator/modules/slurm/templates/slurm_epilog_cm.yaml.tftpl
+++ b/soperator/modules/slurm/templates/slurm_epilog_cm.yaml.tftpl
@@ -4,26 +4,36 @@ resources:
     metadata:
       name: slurm-epilog
     data:
+      # language=bash
       epilog.sh: |-
         #!/bin/bash
 
-        if [ -n "$SLURM_JOB_GPUS" ]; then
-          cd /tmp
-
-          echo "Execute nvidia-smi health check"
-          rm /tmp/nvidia-smi.out 2> /dev/null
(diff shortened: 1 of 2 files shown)


theyoprst (approved):

asteny (approved):

dstaroff:
We decided that we proceed with the previous state of these checks despite their implementation is not recommended by Slurm.
It would be inconvenient to have just `Prolog failed` reason in case of problems. And the mechanism of Slurm does not guarantee the node will be 100% drained in case of non-zero return code.

PR #284, opened 2025-04-10
Add customization for slurm healthchecks to terraform

ChessProfessor:
Supported in soperator here: https://github.com/nebius/soperator/pull/673

### DIFF
--- a/soperator/modules/slurm/templates/helm_values/slurm_cluster.yaml.tftpl
+++ b/soperator/modules/slurm/templates/helm_values/slurm_cluster.yaml.tftpl
@@ -19,6 +19,14 @@ workerFeatures:
     nodesetName: "${feature.nodeset_name}"
   %{~ endfor ~}
 
+healthCheckConfig:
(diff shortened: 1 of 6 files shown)


soperator/modules/slurm/templates/helm_values/slurm_cluster.yaml.tftpl
```diff
@@ -19,6 +19,14 @@ workerFeatures:
     nodesetName: "${feature.nodeset_name}"
   %{~ endfor ~}
 
+healthCheckConfig:
```
theyoprst, line 22: Should we add a check that slurm_health_check_config is not nil as well?

ChessProfessor: Sounds reasonable, let me see

ChessProfessor: Fixed

theyoprst (approved):

PR #313, opened 2025-05-07
cleanup enroot containers in prolog

itechdima:


soperator/modules/slurm/templates/slurm_prolog_cm.yaml.tftpl
```diff
@@ -7,6 +7,14 @@ resources:
       prolog.sh: |-
         #!/bin/bash
 
+        echo "Cleanup enroot containers if it's requeue and there're leftovers of previous run"
```
rdjjke, line 10: What about rephrasing it a little?
> Cleanup leftover enroot containers if the job is restarted with the same ID

rdjjke (dismissed):

soperator/modules/slurm/templates/slurm_prolog_cm.yaml.tftpl
itechdima: done

theyoprst (approved):

PR #319, opened 2025-05-14
🇺🇸Add Kansas City, USA (us-central1) to Soperator Regions

brianlechthaler:


alex000kim:
@brianlechthaler I think `soperator/modules/available_resources/platform.tf` also needs to be updated.

PR #330, opened 2025-05-21
Make Grafana password sensitive in soprator

keyvaann:


theyoprst:
There are conflicts, and it seems that we removed the password in the first place, so I doesn't make sense anymore to declare it as sensitive. Closing

PR #339, opened 2025-05-28
Add scripts folder with bash and python scripts for IB topology conversion

iofits:


jadnov (approved):
Example scripts to get IB topology,
not run by any current jobs.

PR #351, opened 2025-06-11
[FIX] Add buckets cleanup

roman-iurkov:
Due to the some leftovers in the buckets after tests or failed tests, we need to cleanup buckets first

```
Deleting all storage bucket
Error: rpc error: code = InvalidArgument desc = BucketNotEmpty: The bucket is not empty
request = 9b43f813-8dd9-4624-abc1-ab0329d3d3c6
caused by service error:
  Service storage error BucketNotEmpty
```


malibora (approved):
Nice!

PR #356, opened 2025-06-16
Issue 905/fix k8s version

HoomanRamezani:
Avoid hardcoding k8s version
- Setting as null resolves with Nebius' backend default Kubernetes version
- 1.30 will be deprecated end of June

### DIFF
--- a/k8s-training/variables.tf
+++ b/k8s-training/variables.tf
@@ -19,11 +19,11 @@ variable "region" {
   type        = string
 }
 
-# K8s cluster
+# K8s cluster 
 variable "k8s_version" {
-  description = "Kubernetes version to be used in the cluster."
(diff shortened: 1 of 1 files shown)


k8s-training/variables.tf
```diff
@@ -19,11 +19,12 @@ variable "region" {
   type        = string
 }
 
-# K8s cluster
+# K8s cluster -> Keep commented out to use backend default (reccomended)
+#     if you need to choose manually, choose 1.31 or above
```
aaroniscode, line 23: Your comment says to "keep commented out" but it's not commented out, it's set to null. I think you want to explain that setting to null uses the backend default.

aaroniscode (commented):
Left a comment.  Also I think we need a separate PR for Soperator as that is maintained off a different branch, not main.

malibora (approved):

PR #358, opened 2025-06-17
Unhardcode kubernetes version.

theyoprst:


dstaroff (changes requested):
Could you then please mark them with
```hcl
type = optional(string)
nullable = true
```
?

dstaroff (approved):

PR #449, opened 2025-07-21
Rest ensure jail

theyoprst:
- Add custom init container ensuring that jail has right fs type
- Fix vmagent resources

### DIFF
--- a/soperator/modules/slurm/variables.tf
+++ b/soperator/modules/slurm/variables.tf
@@ -495,8 +495,8 @@ variable "resources_vm_agent" {
     cpu    = string
   })
   default = {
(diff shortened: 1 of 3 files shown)


soperator/modules/slurm/variables.tf
```diff
@@ -495,8 +495,8 @@ variable "resources_vm_agent" {
     cpu    = string
   })
   default = {
-    memory = "4Gi"
-    cpu    = "2000m"
+    memory = "10Gi"
+    cpu    = "5000m"
```
Uburro, line 499: maybe add gomaxprocs?

PR #476, opened 2025-08-06
add vars controller_state_on_filestore for controller pool

Uburro:


soperator/modules/slurm/templates/helm_values/terraform_fluxcd_values.yaml.tftpl
```diff
@@ -489,6 +489,17 @@ resources:
                   spool:
                     volumeClaimTemplateSpec: null
                     volumeSourceName: worker-spool
```
rdjjke, line 491: This
```
volumeClaimTemplateSpec: null
volumeSourceName: worker-spool
```
should not be set if use_controller_pvc == false

PR #566, opened 2025-09-23
disk sizes normalized

pbutler:
Normalized references to disk sizes that were in bytes, so that they are multiplied by 1024 the correct number of times to get a human readable value.

- normalized disk size representations in several locations to be more human readable


pbutler:
Looks like because I touched files in certain directories that hadn't been linted before it called out some issues with format.  I'll look into this, confirm, and make the appropriate changes and resubmit.

PR #613, opened 2025-11-13
SCHED-378: Separate node group for NFS

ali-sattari:

Creates a separate node group for NFS in K8S.

### DIFF
--- a/soperator/modules/slurm/locals.tf
+++ b/soperator/modules/slurm/locals.tf
@@ -136,9 +136,12 @@ locals {
       memory = 0.5
     }
     nfs_server = {
+      limits = {
+        memory = var.resources.nfs != null ? var.resources.nfs.memory_gibibytes : 1
+      }
       requests = {
-        memory = 1
-        cpu    = 1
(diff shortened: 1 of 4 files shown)


soperator/modules/slurm/locals.tf
```diff
@@ -131,11 +136,8 @@ locals {
       memory = 0.5
     }
     nfs_server = {
-      limits = {
-        memory = 2
-      }
       requests = {
-        memory = 0.5
+        memory = 1
```
Uburro: by default preset = "32vcpu-128gb" but we set just 1 CPU and Memory. Maybe we can increase this value by default?

Uburro: The higher the CPU requests, the more cpu.shares the CFS scheduler allocates to the process, giving it higher priority when competing for CPU time. Higher requests = more shares = larger proportion of CPU during contention with daemonset pods. 

https://github.com/kubernetes/kubernetes/blob/854e67bb51e177b4b9c012928d8271704e9cb80d/pkg/kubelet/cm/helpers_linux.go#L90-L106

Up2you but INO better set Higher requests for cpu and limit and set limit only for memory

PR #618, opened 2025-11-18
disabling device-plugin for driverfull image

timothy-le7:
This is a small fix. For context, driverfull images already have NVIDIA GPU and networking drivers; they should not have additional drivers/operators.

Prior to this change, an MK8s clusters with driverfull image would additionally try to install device plugins and fail since the driverfull image already has plugins. With this change, driverfull images do not install unncessary plugins and succeed when `gpu_nodes_driverfull_image = true` in `k8s-training/terraform.tfvars`

Huge thanks to @aaroniscode for helping debug!


timothy-le7:
The job failed to an out-of-date image that is being updated in: https://github.com/nebius/nebius-solutions-library/pull/621

PR #889, opened 2026-03-19
Feature/binpacking

pbutler:
This requires merge of #888 feature/opa_gatekeeper.

Add a k8s training module to enable a bin packing scheduler. This allows the k8s scheduling system to prioritize utilization and reduce fragmentation of resources.

### DIFF
--- a/k8s-training/applications.tf
+++ b/k8s-training/applications.tf
@@ -59,17 +59,34 @@ resource "terraform_data" "binpacking_scheduler_version" {
       condition     = local.binpacking_kube_sched_ver != null
       error_message = "When binpacking_enable is true, set k8s_version to a supported minor version or set binpacking_kube_sched_ver to a full kube-scheduler patch version."
     }
+
+    precondition {
+      condition     = !local.binpacking_enable_mutator || var.opa_gatekeeper_enable
+      error_message = "binpacking_forced_namespaces requires opa_gatekeeper_enable = true. Set opa_gatekeeper_enable = true, or set binpacking_forced_namespaces = [] and opt pods in with spec.schedulerName."
+    }
   }
 }
 
 module "binpacking_scheduler" {
-  source             = "../modules/binpacking"
-  count              = var.binpacking_enable ? 1 : 0
-  enable_mutator     = local.binpacking_enable_mutator
-  kube_sched_ver     = local.binpacking_kube_sched_ver
-  mutated_namespaces = var.binpacking_forced_namespaces
+  source         = "../modules/binpacking"
+  count          = var.binpacking_enable && local.binpacking_kube_sched_ver != null ? 1 : 0
+  kube_sched_ver = local.binpacking_kube_sched_ver
+}
+
+resource "kubectl_manifest" "binpacking_mutator" {
+  count = (
+    var.binpacking_enable &&
+    local.binpacking_enable_mutator &&
+    local.binpacking_kube_sched_ver != null &&
+    var.opa_gatekeeper_enable
+  ) ? 1 : 0
+
+  yaml_body = templatefile("../modules/binpacking/files/opa_gatekeeper_mutator.yaml.tftpl", {
+    namespaces = var.binpacking_forced_namespaces
+  })
 
   depends_on = [
-    terraform_data.binpacking_scheduler_version
+    module.opa_gatekeeper,
+    module.binpacking_scheduler,
   ]
 }
--- a/k8s-training/applications.tf
+++ b/k8s-training/applications.tf
@@ -48,11 +48,13 @@ module "kuberay-service" {
 
 module "opa_gatekeeper" {
   source = "../modules/opa_gatekeeper"
-  count  = var.enable_opa_gatekeeper ? 1 : 0
+  count  = var.opa_gatekeeper_enable ? 1 : 0
 }
 
 module "binpacking_scheduler" {
-  source         = "../modules/binpacking"
-  count          = var.enable_binpacking ? 1 : 0
-  enable_mutator = var.force_binpacking_default
+  source             = "../modules/binpacking"
+  count              = var.binpacking_enable ? 1 : 0
+  enable_mutator     = local.binpacking_enable_mutator
+  kube_sched_ver     = local.binpacking_kube_sched_ver
+  mutated_namespaces = var.binpacking_forced_namespaces
 }
--- a/k8s-training/applications.tf
+++ b/k8s-training/applications.tf
@@ -77,7 +77,6 @@ resource "kubectl_manifest" "binpacking_mutator" {
   count = (
     var.binpacking_enable &&
     local.binpacking_enable_mutator &&
-    local.binpacking_kube_sched_ver != null &&
     var.opa_gatekeeper_enable
   ) ? 1 : 0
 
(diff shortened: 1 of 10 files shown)


dasbatta (approved):
@pbutler OPA plugin config looks good and straight forward. I think this workflow will trigger for the Terraform k8s tests. Let's keep an eye. 
@aaroniscode Can you take a look as well.

k8s-training/variables.tf
```diff
@@ -429,3 +429,21 @@ variable "enable_opa_gatekeeper" {
   type        = bool
   default     = false
 }
+
+variable "enable_opa_gatekeeper" {
+  description = "Enable OPA Gatekeeper"
+  type        = bool
+  default     = false
+}
```
aaroniscode, line 512: @pbutler you've got this variable declared twice `enable_opa_gatekeeper`.

modules/binpacking/files/opa_gatekeeper_mutator.yaml.tftpl
```diff
@@ -0,0 +1,19 @@
...
+  applyTo:
+    - groups: [""]
+      kinds: ["Pod"]
+      versions: ["v1"]
+  match:
+    scope: Namespaced
+    kinds:
+      - apiGroups: ["*"]
+        kinds: ["Pod"]
+    namespaces: ${namespaces}
```
aaroniscode: Getting error
```
Call to function "templatefile" failed: ../modules/binpacking/files/opa_gatekeeper_mutator.yaml.tftpl:15,19-29: Invalid template
│ interpolation value; Cannot include the given value in a string template: string required, but have list of string..
```
Codex recommends `jsonencode(namespaces)`

modules/binpacking/files/bp-scheduler.yaml
```diff
@@ -0,0 +1,126 @@
...
+        component: scheduler
+        tier: control-plane
+        version: second
+    spec:
+      serviceAccountName: bp-scheduler
+      containers:
+        - command:
+            - /usr/local/bin/kube-scheduler
+            - --config=/etc/kubernetes/bp-scheduler/bp-scheduler-config.yaml
+          image: registry.k8s.io/kube-scheduler:v1.32.9
```
aaroniscode: You've hard coded `registry.k8s.io/kube-scheduler:v1.32.9` while `k8s-training` allows `k8s_version` to be other supported versions. Kubernetes requires `kube-scheduler` to not be newer than the API server and generally to match the API server minor version, so this can break clusters on 1.31 or future defaults. Can we make the scheduler tag configurable or derive it from the chosen cluster version?

aaroniscode (changes requested):
Hi Patrick. Really love the feature!  I've found a few items for you to review. Look forward to getting this merged.

k8s-training/variables.tf
pbutler: I think this is a result of my fixing a rebase incorrectly :/

modules/binpacking/files/bp-scheduler.yaml
pbutler: Done.  Although if the default is selected for k8s_version (is null) then it has to guess and goes with the 1.32.9 version which should be okay in most circumstances

modules/binpacking/files/opa_gatekeeper_mutator.yaml.tftpl
pbutler: Makes sense, done

k8s-training/variables.tf
pbutler: Fixed and reorganized the variable nameinga bit so they're more clean and clear

dasbatta (approved):
Approving it. Resolved all pending comments and Questions

k8s-training/applications.tf
```diff
@@ -48,5 +48,45 @@ module "kuberay-service" {
...
 module "opa_gatekeeper" {
   source = "../modules/opa_gatekeeper"
-  count  = var.enable_opa_gatekeeper ? 1 : 0
+  count  = var.opa_gatekeeper_enable ? 1 : 0
+}
+
+resource "terraform_data" "binpacking_scheduler_version" {
+  count = var.binpacking_enable ? 1 : 0
+
+  lifecycle {
```
pbutler, line 57: Is there an advantage of doing lifecycle based checking vs variables based checking?

k8s-training/applications.tf
```diff
@@ -48,5 +48,45 @@ module "kuberay-service" {
...
+  source         = "../modules/binpacking"
+  count          = var.binpacking_enable && local.binpacking_kube_sched_ver != null ? 1 : 0
+  kube_sched_ver = local.binpacking_kube_sched_ver
+}
+
+resource "kubectl_manifest" "binpacking_mutator" {
+  count = (
+    var.binpacking_enable &&
+    local.binpacking_enable_mutator &&
+    local.binpacking_kube_sched_ver != null &&
```
pbutler: Leave out the version check because it is already checked for elsewhere.

k8s-training/applications.tf
```diff
@@ -48,5 +48,45 @@ module "kuberay-service" {
...
+    precondition {
+      condition     = !local.binpacking_enable_mutator || var.opa_gatekeeper_enable
+      error_message = "binpacking_forced_namespaces requires opa_gatekeeper_enable = true. Set opa_gatekeeper_enable = true, or set binpacking_forced_namespaces = [] and opt pods in with spec.schedulerName."
+    }
+  }
+}
+
+module "binpacking_scheduler" {
+  source         = "../modules/binpacking"
+  count          = var.binpacking_enable && local.binpacking_kube_sched_ver != null ? 1 : 0
```
pbutler, line 72: I don't love that this will silently turn off binpacking scheduler if user enables it but there is no version set.  This feels like it should be an error condition.

k8s-training/locals.tf
```diff
@@ -76,6 +76,16 @@ locals {
     "gpu-b300-sxm"   = ["all-disabled", "all-enabled", "all-balanced", "all-1g.23gb", "all-1g.23gb.me", "all-1g.45gb", "all-2g.45gb", "all-3g.90gb", "all-4g.90gb", "all-7g.180gb"]
     "gpu-rtx6000"    = ["all-disabled", "all-enabled", "all-balanced", "all-1g.24gb", "all-1g.24gb.me", "all-1g.48gb", "all-2g.48gb", "all-4g.96gb"]
   }
+  binpacking_kube_sched_ver_by_k8s_minor = {
```
pbutler, line 164: Add a comment to say when these values were last updated and that they should be updated regularly.

k8s-training/locals.tf
```diff
@@ -76,6 +76,16 @@ locals {
...
+    "1.30" = "1.30.14"
+    "1.31" = "1.31.14"
+    "1.32" = "1.32.13"
+    "1.33" = "1.33.13"
+    "1.34" = "1.34.9"
+  }
+  binpacking_k8s_minor      = var.k8s_version == null ? null : replace(tostring(var.k8s_version), "/^v?([0-9]+\\.[0-9]+)(?:\\.[0-9]+)?$/", "$1")
```
pbutler: I think we should only rely on the map if the version string only goes to the minor level (1.X) if it goes through to the patch level (1.X.Y) we already know which version is needed and we should use 1.X.Y

aaroniscode (approved):
We are GOOD to go!

PR #907, opened 2026-04-08
SOperator installation example should use no expiring tokens

faucct:

Now there is no need to rerun `.envrc` just to refresh tokens.

### DIFF
--- a/soperator/README.md
+++ b/soperator/README.md
@@ -65,7 +65,7 @@ This command loads environment variables and performs several important setup ta
 - Configures Object Storage access for the Terraform state
 - Exports environment variables with resource IDs 
 
-Check that NEBIUS_IAM_TOKEN is valid:
+Check that nebius CLI is authenticated:
 ```bash
 nebius iam whoami
 ```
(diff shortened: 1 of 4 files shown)


theyoprst (commented):
nice cleanup, but I think the flux provider config is broken:

```hcl
provider "flux" {
  kubernetes = local.kubernetes_exec
}
```

`local.kubernetes_exec` is just `{api_version, command, args}` — the `fluxcd/flux` provider's `kubernetes` attribute expects `host`, `cluster_ca_certificate`, and an `exec = {...}` sub-object. HCL silently drops unknown top-level keys at that layer, which is why `terraform validate` is happy, but at apply time flux ends up with no host/CA and no exec plugin. Did you actually run `terraform apply` with this? I'd expect it to blow up.

I think you want:

```hcl
provider "flux" {
  kubernetes = {
    host                   = module.k8s.control_plane.public_endpoint
    cluster_ca_certificate = module.k8s.control_plane.cluster_ca_certificate
    exec                   = local.kubernetes_exec
  }
}
```

Two other small things:
1. `variables.tf:17` still defines `variable "iam_token"` — nothing consumes it after this PR, should be removed
2. `soperator/README.md:63` ("exports IAM token") and `L68` ("Check that `NEBIUS_IAM_TOKEN` is valid") are now stale

kubernetes and helm provider changes look good to me.

theyoprst (approved):
LGTM, thanks for the fixes

PR #908, opened 2026-04-08
Add reusable Tailscale and External Secrets modules

aaronbfagan:

This PR adds a reusable Terraform module set for private service exposure over Tailscale on Nebius Kubernetes clusters, along with a reusable Nebius MysteryBox + External Secrets integration for secret-backed operator auth.

New modules:

- `modules/tailscale-operator`
- `modules/tailscale-service`
- `modules/external-secrets-operator`
- `modules/external-secret-mysterybox`

### Tailscale operator and service modules

- installs the Tailscale Kubernetes Operator once per cluster
- exposes arbitrary Kubernetes Services privately to a tailnet through operator-managed proxy Services
- supports explicit secret-backed operator auth via `oauth_secret_name`
- defaults `enable_cilium_bpf_lb_sock_hostns_only = true` for Nebius MK8s compatibility
- includes an opt-in workaround for upstream stale proxy behavior:
  - `restart_generated_proxy_once_after_create = true`
  - `restart_generated_proxy_strategy = "template_annotations"` by default
  - `local_exec` remains available as an explicit fallback

### External Secrets modules

- installs External Secrets Operator with Helm
- syncs one Nebius MysteryBox secret into one Kubernetes Secret
- keeps downstream Terraform consumers on a secret-reference contract rather than secret values
- supports both `SecretStore` and `ClusterSecretStore`
- supports either following the MysteryBox primary version or pinning a specific version

### `soperator`

Validated authenticated Slurm REST access over Tailscale:
- operator installed successfully
- unauthenticated request returned `401`
- authenticated `/slurm/v0.0.41/diag` returned valid Slurm JSON

### `k8s-training`

Validated generic arbitrary-service exposure over Tailscale:
- `whoami` served successfully over MagicDNS
- 2048 demo app served successfully over MagicDNS

### Secret-backed auth

Validated end-to-end:
- MysteryBox secret
- External Secrets sync
- Kubernetes Secret
- `tailscale-operator` consuming that Secret via `oauth_secret_name`

### OAuth rotation

Validated fresh-create behavior after rotation:
- promoted a new MysteryBox secret version to primary
- confirmed fresh Tailscale-backed service creation on both `k8s-training` and `soperator`
- after revoking the old credential, fresh service creation still succeeded

### External Secrets bootstrap auth

The Nebius MysteryBox provider in External Secrets currently documents only secret-backed auth methods, so the bootstrap Kubernetes Secret containing Nebius Subject Credentials JSON still has to be created out of band.

### Rotation cleanup caveat

Fresh service creation after credential rotation works, but cleanup of older proxy-backed Services can fail if the old Tailscale OAuth credential is revoked before those older Services are deleted. In testing, this showed up as operator cleanup failures with `401 Unauthorized` / `API token invalid`.

Recommended safe order:
1. create new credential
2. promote new MysteryBox version
3. validate fresh service creation
4. delete any older proxy-backed Services that still need cleanup
5. revoke old credential

## Docs

Included README documentation for all four modules covering:
- design and creation boundaries
- supported deployment patterns
- Nebius/Cilium considerations
- External Secrets bootstrap requirements
- validation guidance
- rotation and troubleshooting notes

### DIFF
--- a/k8s-training/variables.tf
+++ b/k8s-training/variables.tf
@@ -242,7 +242,7 @@ variable "mig_parted_config" {
   default     = null
 
   validation {
-    condition = var.mig_parted_config == null || contains(
+    condition = var.mig_parted_config == null ? true : contains(
       lookup(local.valid_mig_parted_configs, local.gpu_nodes_platform, []),
       var.mig_parted_config,
     )
--- a/modules/external-secret-mysterybox/README.md
+++ b/modules/external-secret-mysterybox/README.md
@@ -1,0 +1,229 @@
+#####################################################################
+# NOTE: This is a module and should not be run manually or standalone
+#####################################################################
+
+Creates an External Secrets `SecretStore` or `ClusterSecretStore` for Nebius
+MysteryBox and an `ExternalSecret` that syncs one MysteryBox secret into a
+Kubernetes Secret.
+
+Use this module when Terraform-managed workloads need a Kubernetes Secret
+reference but should not read the underlying secret value into Terraform
+variables or state.
+
+What this module manages:
+
+- An optional namespace for the ExternalSecret target Secret
+- An optional `SecretStore` or `ClusterSecretStore` for the Nebius MysteryBox
+  provider
+- One `ExternalSecret` that extracts all key/value pairs from one MysteryBox
+  secret into one Kubernetes Secret
+
+What this module does not manage:
+
+- Nebius service account creation
+- Nebius MysteryBox secret creation
+- External Secrets Operator installation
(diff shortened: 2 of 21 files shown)


dasbatta (approved):
Thanks for working on this integration. ESO could be generalized solution as we discussed, but it makes sense for your use case

pbutler (approved):
Looks good to me!  Well documented and clear code.

aaroniscode (approved):
Thanks for the PR @aaronbfagan .  The `k8s-training` test is not passing due to limited capacity so I'll merge without passing tests.

PR #954, opened 2026-05-01
Expose k8s training RTX and Filestore deletion protection options

saurabh22patel:

Add enable_gpu_cluster to the k8s-training tfvars sample so RTX6000 and single-node deployments can explicitly disable GPU clustering while the default 8-GPU training path remains enabled.

Add filestore_forbid_deletion to k8s-training and pass it through to the Nebius filesystem forbid_deletion field, allowing option to protect the shared filestore.

### DIFF
--- a/k8s-training/terraform.tfvars
+++ b/k8s-training/terraform.tfvars
@@ -37,8 +37,7 @@ cpu_nodes_preset   = "4vcpu-16gb" # CPU nodes preset
 gpu_nodes_platform = "gpu-h200-sxm"        # GPU nodes platform: gpu-h100-sxm, gpu-h200-sxm, gpu-b200-sxm
 gpu_nodes_preset   = "8gpu-128vcpu-1600gb" # GPU nodes preset: 8gpu-128vcpu-1600gb, 8gpu-128vcpu-1600gb, 8gpu-160vcpu-1792gb
 # Infiniband fabrics: https://docs.nebius.com/compute/clusters/gpu#fabrics
-infiniband_fabric  = ""   # Infiniband fabric name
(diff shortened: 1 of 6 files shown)


k8s-training/terraform.tfvars
```diff
@@ -37,7 +37,8 @@ cpu_nodes_preset   = "4vcpu-16gb" # CPU nodes preset
 gpu_nodes_platform = "gpu-h200-sxm"        # GPU nodes platform: gpu-h100-sxm, gpu-h200-sxm, gpu-b200-sxm
 gpu_nodes_preset   = "8gpu-128vcpu-1600gb" # GPU nodes preset: 8gpu-128vcpu-1600gb, 8gpu-128vcpu-1600gb, 8gpu-160vcpu-1792gb
 # Infiniband fabrics: https://docs.nebius.com/compute/clusters/gpu#fabrics
-infiniband_fabric = "" # Infiniband fabric name
+infiniband_fabric  = ""   # Infiniband fabric name
+enable_gpu_cluster = true # Disable for RTX6000 deployments or single-node deployments.
```
pbutler: I know you're just explicitly setting a variable here that someone else created, but should we consider that if infiniband_fabric is empty or null that enable_gpu_cluster should necessarily be false instead of having to set two different variables?

pbutler (approved):
I think this looks good.  I just want you to consider my one comment and if it makes sense to do it a slightly different way to reduce the number of needed variables.  I would say it's merge-able as is though.

Thanks!

k8s-training/terraform.tfvars
saurabh22patel: Good input, i will update to make things more streamlined here.

aaroniscode (approved):

PR #955, opened 2026-05-02
custom-nat-gateway: add existing-VPC NAT example

realvz:
## What changed

- adds a new `custom-nat-gateway` Terraform example
- creates a gateway subnet, a private workload subnet, a custom route table, and a gateway VM that handles outbound NAT
- expects an existing VPC through `vpc_network_id` instead of creating a new network
- adds a CLI-backed validation step so the run fails early if the VPC is missing or belongs to the wrong project
- adds `deploy_test_vm` so the private workload VM can be turned off when you only want the gateway path
- adds README and smoke-test coverage for the example

## Validation

- `terraform -chdir=custom-nat-gateway fmt`
- `terraform -chdir=custom-nat-gateway init -backend=false`
- `terraform -chdir=custom-nat-gateway validate`
- `terraform -chdir=custom-nat-gateway test` against a live Nebius project

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,10 @@ Enhance security with a Wireguard VPN instance by minimizing the use of public I
 
 Deploys a Bastion instance that serves as a secure jump host for your infrastructure. It improves the security by minimizing the use of Public IPs and limiting access to the rest of the environment. 
 
+[Custom NAT gateway](./custom-nat-gateway/README.md)
+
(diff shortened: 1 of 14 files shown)


dasbatta (approved):
It follows the TF recipe for Public docs
https://docs.nebius.com/vpc/routing/custom-nat-gateway

PR #958, opened 2026-05-05
Harden k8s-training CI failure handling and bucket cleanup

aaronbfagan:
This PR hardens the `k8s-training` Terraform CI path so mk8s/GPU provisioning failures are detected more clearly, bounded by timeout, and followed by more reliable cleanup.

It also fixes a stale KubeRay test input and a Terraform test report upload path issue that could otherwise make CI fail for the wrong reason.

### Fix stale KubeRay test input

Updated `k8s-training/tests/k8s-training-kuberay.tftest.hcl`.

Changed stale input:

`enable_kuberay = true`

to the currently declared variable:

`enable_kuberay_cluster = true`

The old `enable_kuberay` variable is no longer declared by `k8s-training`. The current variables are `enable_kuberay_cluster` and `enable_kuberay_service`.

### Add k8s-training Terraform Test watcher

Updated `.github/workflows/terraform.yml` so non-`k8s-training` solutions still run plain `terraform test`.

For `k8s-training`, the workflow now runs `terraform test` in the background and watches the mk8s GPU node group for known fatal provisioning events.

The watcher fails early on:

- `ComputeInstanceCreationFailed`
- `NodeNotFound` with message containing `STOPPED for longer than`

This does not make GPU capacity available or make mk8s provisioning succeed. It makes known terminal provisioning failures visible earlier instead of letting CI hang or fail later with less useful context.

### Add Terraform Test timeout

Added a hard upper bound to the `Terraform Test` step:

`timeout-minutes: 150`

If the watcher does not catch a known fatal event, the step is still bounded.

### Improve bucket cleanup

Updated `cleanup-infra` bucket cleanup behavior:

- Removed storage buckets from the generic forced cleanup loop.
- Added Nebius-native bucket deletion with `nebius storage bucket delete --ttl 0`.
- Added best-effort purge for buckets with `status.purge_at`.
- Continued excluding `terraform-test-reports`.

This is intended to handle versioned bucket cleanup more reliably than recursive `s3cmd del`, especially when object versions or delete markers remain.

### Fix Terraform report upload path

Changed the report upload path from:

`${{ matrix.solution.name }}`

to:

`${{ matrix.solution }}`

`matrix.solution` is a string in this workflow, not an object.

### Formatting

Applied a small `terraform fmt` alignment change in `k8s-training/tests/main.tftest.hcl`.

### Git and mergeability

Ran:

`git fetch origin`

`git merge-tree --write-tree origin/main HEAD`

`git diff --check origin/main...HEAD`

Result: no merge conflicts and no whitespace issues.

### Workflow syntax

Validated that the GitHub Actions workflow YAML parses successfully.

Also checked embedded workflow `run:` blocks with `bash -n` after substituting GitHub expressions.

Result: passed.

### Terraform formatting

Ran from `k8s-training`:

`terraform fmt -check -recursive . ../modules`

Result: passed.

### Terraform validate

Ran `terraform init -backend=false` and `terraform validate` in a detached temporary worktree to avoid polluting the real checkout.

Result: passed.

### Mock workflow harness

Ran a local mock harness against the actual workflow snippets to verify behavior without creating Nebius resources.

Covered cases:

- Non-`k8s-training` still runs plain `terraform test`.
- `k8s-training` succeeds when no fatal node-group event is returned.
- `k8s-training` exits `1` when a mocked `NodeNotFound` / `STOPPED for longer than` event appears.
- Report upload path uses `${{ matrix.solution }}`.
- Bucket delete skips `terraform-test-reports` and uses `delete --ttl 0`.
- Bucket purge only purges non-report buckets with `status.purge_at`.

Result: passed.

## Notes

This PR intentionally does not run the real `terraform test` path locally, because that requires the CI Nebius project, credentials, and mk8s/GPU resources. The real GitHub Actions run is still the authoritative test for actual mk8s/GPU provisioning behavior.

### DIFF
--- a/.github/workflows/terraform.yml
+++ b/.github/workflows/terraform.yml
@@ -180,7 +180,78 @@ jobs:
 
     # Run Terraform Tests
     - name: Terraform Test
-      run: terraform test -junit-xml=tests/reports/TEST-result-${{ github.run_id }}.xml
+      timeout-minutes: 150
+      run: |
+        report_path="tests/reports/TEST-result-${{ github.run_id }}.xml"
+
+        if [ "${{ matrix.solution }}" != "k8s-training" ]; then
+          terraform test -junit-xml="${report_path}"
+          exit $?
+        fi
+
+        cluster_name="${TF_VAR_cluster_name:-k8s-training}"
+        gpu_node_group_name="${cluster_name}-ng-gpu-0"
+        poll_seconds=60
+
+        terraform test -junit-xml="${report_path}" &
+        test_pid=$!
+
+        stop_test() {
+          if kill -0 "${test_pid}" 2>/dev/null; then
+            kill "${test_pid}" || true
+            wait "${test_pid}" || true
+          fi
+        }
+
+        # Fail early when the GPU node group reaches a terminal provisioning state.
+        while kill -0 "${test_pid}" 2>/dev/null; do
+          cluster_id="$(
+            { nebius mk8s v1 cluster list --parent-id "${TF_VAR_parent_id}" --format json 2>/dev/null || true; } \
+              | jq -r --arg name "${cluster_name}" '
+                  [(.items // [])[] | select(.metadata.name == $name)]
+                  | sort_by(.metadata.created_at)
+                  | last
+                  | .metadata.id // empty
+                '
+          )"
+
+          if [ -n "${cluster_id}" ]; then
+            node_group_json="$(
+              { nebius mk8s v1 node-group list --parent-id "${cluster_id}" --format json 2>/dev/null || true; } \
+                | jq -c --arg name "${gpu_node_group_name}" '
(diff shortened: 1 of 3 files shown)


dasbatta (approved):
LGTM. Approving, let other check as well. 

As discussed on slack, 
There will be a diff ticket to address, cluster polling conflict. 
"since this PR adds polling for the k8s-training CI run, can we also scope that polling to the specific CI deployment and cluster identity.
Right now CI uses the default cluster name 'k8s-training', and the poller resolves the cluster by name itself.
Could we either:
1. set a run-specific 'TF_VAR_cluster_name' before init/plan/test, or
2. have the poller read the exact cluster id from Terraform state/output instead of resolving by name?

aaroniscode (approved):
Thank you @aaronbfagan for making the CI better!

PR #959, opened 2026-05-06
SCHED-292: Configure OpenTelemetry batch settings

ChessProfessor:

Adds Terraform-level configuration for OpenTelemetry batch processor settings used by logs, jail logs, and events collectors. By default, Terraform does not override the chart values; users can set opentelemetry_batch to customize timeout, send_batch_size, and send_batch_max_size when needed.

### DIFF
--- a/soperator/installations/example/variables.tf
+++ b/soperator/installations/example/variables.tf
@@ -1225,6 +1225,54 @@ variable "kube_state_metrics_max_scrape_size" {
   }
 }
 
+variable "opentelemetry_batch" {
+  description = "OpenTelemetry batch processor overrides for logs, jail logs, and events collectors. Leave null to use chart defaults."
+  type = object({
+    timeout             = optional(string)
+    send_batch_size     = optional(number)
+    send_batch_max_size = optional(number)
+  })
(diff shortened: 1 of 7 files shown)


soperator/installations/example/variables.tf
```diff
@@ -1225,6 +1225,54 @@ variable "kube_state_metrics_max_scrape_size" {
...
+    error_message = "opentelemetry_batch.timeout must be non-empty when set."
+  }
+
+  validation {
+    condition = (
+      var.opentelemetry_batch == null ||
+      var.opentelemetry_batch.send_batch_size == null ||
+      var.opentelemetry_batch.send_batch_size > 0
+    )
+    error_message = "opentelemetry_batch.send_batch_size must be greater than 0 when set."
```
Uburro, line 1253: wjy? helm chat has a default paramerts

ChessProfessor: Sorry but what do you mean? We allow setting null to this field, so default params will be used in such case, this validation only checks that override if exists is positive integer

Uburro (approved):

PR #964, opened 2026-05-07
SCHED-1628: Support adding SSH keys to VMs without public IP addresses

Uburro:

### DIFF
--- a/soperator/installations/example/terraform.tfvars
+++ b/soperator/installations/example/terraform.tfvars
@@ -670,7 +670,6 @@ cleanup_bucket_on_destroy = false
 k8s_version = 1.33
 
 # SSH user credentials for accessing k8s nodes.
-# That option add public ip address to every node.
 # By default, empty list.
 # ---
 # k8s_cluster_node_ssh_access_users = [{
@@ -680,6 +679,11 @@ k8s_version = 1.33
 #     "<ENCRYPTION-METHOD2 HASH2 USER1>",
 #   ]
 # }]
+
+# By default, SSH keys are added without public IP addresses.
+# Set to true to assign public IP addresses to k8s nodes.
+# ---
+# k8s_cluster_node_ssh_access_public_ip = false
 
 # Lines to write to /etc/modprobe.d/nvidia_config.conf via cloud-init (GPU workers only).
 # One option per line.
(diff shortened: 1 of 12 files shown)


soperator/installations/example/terraform.tfvars
```diff
@@ -681,6 +680,11 @@ k8s_version = 1.33
 #   ]
 # }]
 
+# Assign public IP addresses to k8s nodes when k8s_cluster_node_ssh_access_users is configured.
+# Set to false to add SSH keys without public IP addresses.
+# ---
+# k8s_cluster_node_ssh_access_public_ip = true
```
asteny: I think default should be false

asteny (approved):

soperator/installations/example/terraform.tfvars
Uburro: default is false. it's just a comment

theyoprst (commented):
couple of things:

1. default mismatch between layers - `installations/example/variables.tf:505` defaults to `false`, but `modules/k8s/variables.tf:257` defaults to `true`. is this intentional? if it's for backward compat at the module level (so existing direct module consumers don't lose public IPs), worth a comment in the module variable. otherwise the asymmetry will confuse people.

2. docs are misleading about the default. both the README and `terraform.tfvars` say "Set to false to add SSH keys without public IP addresses", which implies the default is `true`. but the actual default in `installations/example/variables.tf` is `false`. the commented example `# k8s_cluster_node_ssh_access_public_ip = true` reinforces the wrong impression - this is the line @asteny already got confused by. suggest:
   - flip the example to `# k8s_cluster_node_ssh_access_public_ip = false` (showing the default), or `# k8s_cluster_node_ssh_access_public_ip = true  # opt-in to public IPs`
   - reword to "Set to true to assign public IP addresses" so it matches the actual default

3. existing users of `installations/example` who already configured `k8s_cluster_node_ssh_access_users` will lose public IPs on next apply (default flipped from effectively-true to false). probably worth calling out in the release notes so people know to set `= true` explicitly if they want to keep current behavior.

minor: the `node_ssh_access_public_ip = { enabled = ... }` local wrapper in `modules/k8s/locals.tf` mimics the `node_ssh_access` pattern but adds nothing - it's just a single bool. could be `local.node_ssh_access_public_ip = local.node_ssh_access.enabled && var.node_ssh_access_public_ip` and used as `local.node_ssh_access_public_ip ? {} : null`. not a blocker.

PR #970, opened 2026-05-11
[FEATURE] Add local NVME disks support

roman-iurkov:
The B300 platform supports local disks, though additional steps are required to enable them within the tenant. This PR was created to add support for deployment via Terraform.

### DIFF
--- a/modules/cloud-init/simple-setup-init.tftpl
+++ b/modules/cloud-init/simple-setup-init.tftpl
@@ -1,4 +1,12 @@
 #cloud-config
+package_update: true
+package_upgrade: true
+%{if enable_local_disks}
+packages:
+  - mdadm
+  - nvme-cli
+
+%{endif}
 users:
 %{ for user in users}
   - name: ${user.user_name}
@@ -9,12 +17,8 @@ users:
       - ${user.ssh_public_key}
 %{ endfor}
 
-
+%{ if extra_disk_id != "" || shared_filesystem_id != "" || aws_access_key_id != "" || enable_local_disks }
 runcmd:
-- apt-get update
-- apt-get upgrade
-
-
 # mount disk if provided
 %{ if extra_disk_id != "" }
 # Prepare partition on secondary disk
@@ -76,8 +80,6 @@ runcmd:
 %{endif}
 
 %{if enable_local_disks}
-# install nvme
-- apt install -y nvme-cli
 # Create RAID 0 array
 - mdadm --create --force --verbose /dev/md0 --level=0 --raid-devices=6 /dev/nvme[0-5]n1
 - mkfs.ext4 /dev/md0
@@ -91,3 +93,4 @@ runcmd:
 - mdadm --detail --scan >> /etc/mdadm/mdadm.conf
 - update-initramfs -u
 %{endif}
+%{ endif }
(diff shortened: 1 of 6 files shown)


modules/instance/main.tf
aaroniscode: @roman-iurkov Boot disk is hard coded to CUDA 12 but B300 requires CUDA 13.  Can change this so it's not hard coded and ideally defaults to right value for each platform type but can be overridden by a variable?

aaroniscode (changes requested):

modules/cloud-init/simple-setup-init.tftpl
```diff
@@ -74,3 +74,20 @@ runcmd:
 - mount-s3 --upload-checksums=off --maximum-throughput-gbps=200 --allow-delete --allow-overwrite --allow-other --endpoint-url=https://storage.eu-north1.nebius.cloud:443 ${mount_bucket} ${s3_mount_path}
 %{endif}
 %{endif}
+
+%{if enable_local_disks}
+# install nvme
+- apt install -y nvme-cli
+# Create RAID 0 array
+- mdadm --create --force --verbose /dev/md0 --level=0 --raid-devices=6 /dev/nvme[0-5]n1
```
pbutler: runcmd only gets run once on first boot.  if the instance is stop/started then it could land on a node with blank nvme's that are un-initialized.

pbutler: Also with nofail this situation would cause everything to stop booting

modules/cloud-init/simple-setup-init.tftpl
```diff
@@ -74,3 +74,20 @@ runcmd:
 - mount-s3 --upload-checksums=off --maximum-throughput-gbps=200 --allow-delete --allow-overwrite --allow-other --endpoint-url=https://storage.eu-north1.nebius.cloud:443 ${mount_bucket} ${s3_mount_path}
 %{endif}
 %{endif}
+
+%{if enable_local_disks}
+# install nvme
+- apt install -y nvme-cli
```
pbutler: use `apt-get install -y` apt complains about an unstable cli

pbutler (changes requested):
A couple of changes to consider.

modules/cloud-init/simple-setup-init.tftpl
dasbatta: This hard-codes exactly six NVMe devices with `/dev/nvme[0-5]n1` and `--raid-devices=6`, but the local disk count depends on the selected platform/preset. Please discover the actual local NVMe devices at boot and pass the detected count to `mdadm` instead of assuming six drives. For example, build a `disks=(...)` array after filtering to the intended local NVMe devices, then use `--raid-devices=${#disks[@]} "${disks[@]}"`.

dasbatta (commented):
Added one inline comment on the hard-coded NVMe drive count.

modules/instance/main.tf
roman-iurkov: @aaroniscode the boot_disk_image variable added with default set to the CUDA13 for each platform. For the simplicity we could keep it like this, as CUDA13 is the recommended image family: https://docs.nebius.com/compute/storage/boot-disk-images#compatibility-types

PR #971, opened 2026-05-12
[FEATURE] Mk8s local NVME support for B300

roman-iurkov:
PR to support local NVME option on B300

### DIFF
--- a/k8s-training/main.tf
+++ b/k8s-training/main.tf
@@ -175,6 +175,9 @@ resource "nebius_mk8s_v1_node_group" "gpu" {
     } : null
 
     local_disks = var.gpu_enable_local_disks ? {
+      config = { 
+        none = true 
+      }
       passthrough_group = {
         requested = true
       }
--- a/k8s-training/main.tf
+++ b/k8s-training/main.tf
@@ -102,10 +102,12 @@ resource "nebius_mk8s_v1_node_group" "cpu-only" {
       priority      = 3
     } : null
     cloud_init_user_data = templatefile("${path.module}/../modules/cloud-init/k8s-cloud-init.tftpl", {
-      enable_filestore     = var.enable_filestore ? "true" : "false",
-      filestore_mount_path = local.filestore.mount_path,
-      ssh_user_name        = var.ssh_user_name,
-      ssh_public_key       = local.ssh_public_key
+      enable_filestore       = var.enable_filestore ? "true" : "false",
+      filestore_mount_path   = local.filestore.mount_path,
+      ssh_user_name          = var.ssh_user_name,
+      ssh_public_key         = local.ssh_public_key
+      gpu_enable_local_disks = false
+      local_nvme_drives_path = "/scratch"
     })
   }
 }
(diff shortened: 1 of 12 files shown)


k8s-training/main.tf
```diff
@@ -172,12 +174,20 @@ resource "nebius_mk8s_v1_node_group" "gpu" {
       priority      = 3
     } : null
 
+    local_disks = var.gpu_enable_local_disks ? {
+      passthrough_group = {
+        requested = true
+      }
+    } : null
+
```
aaroniscode, line 185: @roman-iurkov you must configure `config = { none = true }` otherwise the default according to the docs is set to `kubelet_ephemeral`:
> kubelet_ephemeral: combine all local disks into a single storage volume and use it as kubelet’s local ephemeral storage on the node See also https://kubernetes.io/docs/concepts/storage/ephemeral-storage/
The default when LocalDisksSpecConfig is not set.

Source: https://docs.nebius.com/terraform-provider/reference/resources/mk8s_v1_node_group#nested-schema-for-template-local_disks-config

aaroniscode (changes requested):

dasbatta:
NVMe setup only runs when filestore is enabled: 

@roman-iurkov The local NVMe setup is currently nested under the filestore condition, so `gpu_enable_local_disks=true` does not render any RAID/mount setup when `enable_filestore=false`. Since the default tfvars has filestore disabled, this can request local disks on the node group but never mount `/scratch`.

Pr lines: https://github.com/nebius/nebius-solutions-library/blob/851fda9a6adb51d15b82ce30f92862cd91399c07/modules/cloud-init/k8s-cloud-init.tftpl#L4-L30

The whole runcmd block, including NVMe, is inside:
%{ if enable_filestore != "false" }

modules/cloud-init/k8s-cloud-init.tftpl
```diff
@@ -9,6 +9,23 @@ runcmd:
  - [ modprobe, fuse ]
  - [ mkdir, '-p', ${filestore_mount_path} ]
  - [ mount, '-a' ]
+ %{if gpu_enable_local_disks}
+ # install nvme
+ - apt install -y nvme-cli
```
pbutler: `apt-get install -y`

modules/cloud-init/k8s-cloud-init.tftpl
```diff
@@ -9,6 +9,23 @@ runcmd:
...
+ # Create RAID 0 array
+ - mdadm --create --force --verbose /dev/md0 --level=0 --raid-devices=6 /dev/nvme[0-5]n1
```
pbutler: Consider something like: 
`disks=(/dev/nvme*n1); mdadm --create /dev/md0 --level=0 --raid-devices=${#disks[@]} "${disks[@]}".`
This fixes the reliance one the number of disks to be 0. 
This also suffers from the same problem as the similar line in #970, as it is only run on first boot.

modules/cloud-init/simple-setup-init.tftpl
```diff
@@ -74,3 +74,20 @@ runcmd:
 - mount-s3 --upload-checksums=off --maximum-throughput-gbps=200 --allow-delete --allow-overwrite --allow-other --endpoint-url=https://storage.eu-north1.nebius.cloud:443 ${mount_bucket} ${s3_mount_path}
 %{endif}
 %{endif}
+
+%{if enable_local_disks}
+# install nvme
+- apt install -y nvme-cli
```
pbutler, line 80: `apt-get install -y`

modules/cloud-init/simple-setup-init.tftpl
```diff
@@ -74,3 +74,20 @@ runcmd:
...
+# Create RAID 0 array
+- mdadm --create --force --verbose /dev/md0 --level=0 --raid-devices=6 /dev/nvme[0-5]n1
```
pbutler, line 82: same as above

pbutler (changes requested):
Some things to consider.

k8s-training/main.tf
roman-iurkov: The config has been added

modules/cloud-init/k8s-cloud-init.tftpl
roman-iurkov: Refactored the cloud-init config, shouldn't be the issue anymore

aaroniscode (approved):
Thanks for addressing my item. Approved from my side on the narrow item I addressed. As the PR has had some re-architecting, it should get a holistic review from another CSA.

pbutler (approved):
Looks good to me. (discussed a few q's on slack)

modules/cloud-init/k8s-cloud-init.tftpl
cyril-k, line 1: looks like only indentation is changed in cloudinit, is this really necessary?

roman-iurkov: Reverted

cyril-k (approved):

PR #972, opened 2026-05-12
[FIX] Fix DSVM tests

roman-iurkov:
Since we've migrated to the new region, there is no longer this taupe of VM preset available + pipeline fixes


jadnov (approved):
after Roman request for test fixes

PR #975, opened 2026-05-13
SCHED-1573: GB300 Rack Support

ali-sattari:

Adds support for GB300 platform to our recipe for Soperator:

- New gpu-gb300 platform/preset metadata
- CUDA/driver preset mappings
- Slurm GRES and CPU topology
- GB300 worker validation
- Rack-based worker nodeset naming
- NVLink instance group creation

It also changes worker handling so GB300 nodesets are intended to expand into rack-scoped Slurm NodeSets and mk8s node groups named like `<name>-rack-0-worker`.


soperator/installations/example/driver_presets.tf
```diff
@@ -19,6 +20,7 @@ locals {
       gpu-b200-sxm-a = ["cuda12.8", "cuda13.0"]
       gpu-b300-sxm   = ["cuda13.0"]
       gpu-rtx6000    = ["cuda13.0"]
+      gpu-gb300      = ["cuda13.0"]
```
asteny, line 23: As I remember gb200 required 13.1+

dstaroff (approved):

soperator/installations/example/variables.tf
```diff
@@ -810,14 +830,104 @@ variable "slurm_nodeset_workers" {
     node_local_jail_submounts = []
   }]
 
+  validation {
+    condition = alltrue([
+      for worker in var.slurm_nodeset_workers :
+      worker.size > 0 && floor(worker.size) == worker.size
```
asteny: how to downscale with terraform nodeset to 0?

soperator/installations/example/terraform.tfvars
```diff
@@ -323,7 +323,15 @@ slurm_nodeset_controller = {
 
 # Configuration of Slurm Worker node sets.
 # Multiple worker nodesets are supported with different hardware configurations.
-# Each nodeset will be automatically split into node groups of max 100 nodes with autoscaling enabled.
+# Each nodeset will be automatically split into fixed-size node groups.
+# GB300 workers must use size divisible by 18 in production.
+# Non-production GB300 clusters may use one partial rack with size less than 18.
+# Generated GB300 mk8s node groups are rack-sized, with 18 nodes except for the non-production partial-rack case.
+# Their effective nodeset prefixes are generated as <name>-rack-<rack>-worker.
+# For example, a GB300 worker named "primtrain" creates primtrain-rack-0-worker-0..primtrain-rack-0-worker-17
```
rdjjke: But in case of non-GB300 nodesets, "worker" prefix is the name of the nodeset.
Wouldn't it make more sense if GB300 workers were named `<nodeset-name>-rack-<num>-<ordinal>`?
Is the problem here are two subsequent ordinals, i.e.`rack-1-3`?
If so, maybe we can also shorten these names into `<nodeset-name>-r<num>-<ordinal>`? So that an example of a real worker name would be `worker-r1-3`?
Slurm also likes short names for everything so that they always fit into constant-size columns in CLI tools

soperator/installations/example/terraform.tfvars
```diff
@@ -323,7 +323,15 @@ slurm_nodeset_controller = {
 
 # Configuration of Slurm Worker node sets.
 # Multiple worker nodesets are supported with different hardware configurations.
-# Each nodeset will be automatically split into node groups of max 100 nodes with autoscaling enabled.
+# Each nodeset will be automatically split into fixed-size node groups.
+# GB300 workers must use size divisible by 18 in production.
+# Non-production GB300 clusters may use one partial rack with size less than 18.
```
rdjjke, line 333: Are we 100% sure that <18 nvlink instance groups will never be used in production?
There could also be cases when non-production clusters with several partial racks are needed, and such cases won't be configurable via these TF vars.
Can we add one more variable like `nodes_per_rack` (for example inside of the `nvlink` structure)?

soperator/installations/example/terraform.tfvars
```diff
@@ -169,7 +169,7 @@ slurm_nodesets_partitions = [
   {
     name         = "main"
     is_all       = true
-    nodeset_refs = [] # e.g. ["worker"], but is_all must be false in this case
+    nodeset_refs = [] # e.g. ["worker"] or ["primtrain-rack-0-worker"], but is_all must be false in this case
```
rdjjke: The example `primtrain-rack-0-worker` is strange because the variable is named `nodeset_refs`, but `primtrain-rack-0-worker` is not a nodeset name, this is a nodegroup or a Slurm nodeset name (which is not directly configurable). The nodeset name that the user specified below is `primtrain` in this case.
But I don't know how to easily fix this...

Before, TF nodesets were equal to Slurm nodesets, but now they become different entities. One TF nodeset produces several Slurm nodesets under the hood. Maybe we should at least rename this var to slurm_nodeset_refs and mention this pecularity in the comment?

soperator/installations/example/terraform.tfvars
ali-sattari: The current public API (and provider) doesn't even have a `size` parameter. If that changes at some point to allow sub-rack usage, we need to revisit this recipe (variable usage and validation blocks)

soperator/installations/example/driver_presets.tf
ali-sattari: GB300 works with CUDA13.x
We will add support to create a cluster in this (and following) PRs. More specific version for the best performance is under discussion.

soperator/installations/example/terraform.tfvars
ali-sattari: I though of `<name>-rack-#-worker-#` for readability when using `sinfo` and similar commands. We can shorten to `<name>-r-#-w-#` or similar.
The only part that I suggest to keep is to have worker number per rack. To enable using ranges like `rack-[1-3]-worker-[10-17]` if needed. Does that make sense?

soperator/installations/example/variables.tf
ali-sattari: Good point. 
I checked validations and we currently can't scale down worker nodesets to 0 even without this newly added validation, [this one](https://github.com/nebius/nebius-solutions-library/blob/5580ba7ecc13c75453d78d206a431ec7848de40e/soperator/installations/example/variables.tf#L1185) refers to `each.value.min_size` which doesn't exist for worker (and others).
I will see how I can fix this.

soperator/installations/example/terraform.tfvars
ali-sattari: I changed the name and added a comment to make it less confusing.

soperator/installations/example/variables.tf
ali-sattari: Pushed a change, please check again

soperator/installations/example/terraform.tfvars
rdjjke: > The only part that I suggest to keep is to have worker number per rack.
I didn't get the question. But yes any ranges will work with any of these naming approaches.

Regarding your naming suggestion: I just suggest to avoid putting "worker" word there if it's not the nodeset name, otherwise it will be different for GB300 and non-GB300.
In non-GB300 users can rename "workers" to "nodes" or "compute" or "h200" or anything, but in GB300 they won't be able to.

What about `<name>-rack#-#`? The last ordinal is obviously the node ordinal, and the preceding `rack#` will be like what rack it is -- `rack0` or `rack1`, and without `-` so that two numbers look different.

soperator/installations/example/terraform.tfvars
rdjjke: But if we just land a 16-node nodegroup on one rack, won't it be the same?

rdjjke: So, we bind nodegroups to nvl instance groups, right?
And we can control how many nodes to create per nodegroup in each nodeset.
So if we won't make the nodegroup size constant (18), and will make it configurable in the nodeset spec, the problem will be solved, won't it be?

ali-sattari: We can make nodegroup of `size < 18`. However we can't make `nvl_instance_group` of `size < 18`. This means that customer will be billed for the entire rack (18 nodes) regardless of nodegroup size. So in case of one nodeset it doesn't make sense to set the `size < 18`.

In what case might someone want a nodeset (and nodegroup) smaller than 18 (with GB300)?
Are you thinking about a case where customer want to make multiple nodesets that share racks? e.g. a nodeset with  `size=16` and another nodeset with `size=2`, but they share `nvl_instance_group` and therefore rack?

soperator/installations/example/terraform.tfvars
ali-sattari: Understood, I will make changes to have `<name>-rack#-#`

ali-sattari: Pushed a change to reflect this.

soperator/installations/example/terraform.tfvars
rdjjke: OK, I didin't know that nvl_instance_groups have sizes as well.

> customer will be billed for the entire rack (18 nodes) regardless of nodegroup size.

But this can also be fine in some cases. For example when 2 remaining nodes run out of Soperator, as you described.
Or for development when several people need to share the same rack.

What concerns me is that the current solution is not compatible with <18 nodegroups because it has the logic of splitting nodesets into exactly 18-node nodegroups

rdjjke (approved):

PR #977, opened 2026-05-13
Fix bastion public IP allocation

realvz:
## What changed

The bastion solution now passes `public_ip_allocation_id` into the instance public IP config. Before this change, the README described a stable public IP path, but Terraform always requested a fresh public IP.

I also changed the README to tell users to set the value in `terraform.tfvars`, not `variables.tf`.

## Checks

- `terraform -chdir=/Users/realz/code/nebius-solutions-library/bastion fmt -check`
- `terraform -chdir=/Users/realz/code/nebius-solutions-library/bastion validate`
- `git diff --check`


aaroniscode (approved):
Thanks for this bug fix Re!  Discussed in our PR Review call

PR #1000, opened 2026-06-02
[ADD] Use block topology for GB300

dstaroff:
## Release Notes

As for GB300, default topology is set to:
```
TopologyPlugin=topology/block
TopologyParam=BlockAsNodeRank
```
with `topology.conf`'s
```
BlockSizes=18
```

### DIFF
--- a/soperator/installations/example/main.tf
+++ b/soperator/installations/example/main.tf
@@ -608,6 +608,11 @@ module "slurm" {
     }
   }]
 
+  topology = {
+    plugin     = local.gb300_enabled ? "topology/block" : "topology/tree"
+    block_size = local.gb300_enabled ? 9 : null
+  }
+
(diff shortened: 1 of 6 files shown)


soperator/installations/example/main.tf
```diff
@@ -608,6 +608,11 @@ module "slurm" {
     }
   }]
 
+  topology = {
+    plugin     = local.gb300_enabled ? "topology/block" : "topology/tree"
+    block_size = local.gb300_enabled ? 9 : null
```
ali-sattari: Block size is a parameter we might want to change per customer, it make more sense to move it to a variable and expose it in tfvars.

rdjjke:
I think the default block size should be 18, not 9

ali-sattari (approved):

PR #1001, opened 2026-06-03
SCHED-1786 Enabling essential active checks on GB300

asteny:

### DIFF
--- a/soperator/modules/slurm/locals_active_checks.tf
+++ b/soperator/modules/slurm/locals_active_checks.tf
@@ -1,4 +1,37 @@
 locals {
+  gb300_disabled_active_checks = toset([
+    "all-reduce-perf-nccl-in-docker",
+    "all-reduce-perf-nccl-with-ib",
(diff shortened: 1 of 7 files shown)


soperator/modules/slurm/locals_active_checks.tf
```diff
@@ -1,4 +1,37 @@
 locals {
+  gb300_disabled_active_checks = toset([
```
ali-sattari, line 2: Why not have an allow-list of "gb300_enabled_active_checks" instead? It can be more future proof: if we add new active checks we won't need to update this unless it is GB300 compatible. So it will be deliberate opt-in.

asteny: But for allow-list  we need to copy-paste everytime checks from helm to the terraform for enabling or disabling them. Right now that implemented as disable check in terraform. Do you want to change terraform+helm behaviour for that?

ali-sattari (approved):

PR #1043, opened 2026-06-29
SCHED-1900: Use `TF_CLI_ARGS` instead of `TFE_PARALLELISM` for parallel TF execution

dstaroff:
## Release Notes

`TFE_PARALLELISM` is not recognized by the Terraform CLI, as this option is meant for Terraform Enterprise.
Instead, `TF_CLI_ARGS` per `plan`, `apply`, and `destroy` commands are used.

`.envrc` sourcing now logs more info with clear visual separation.

### DIFF
--- a/soperator/installations/example/.envrc
+++ b/soperator/installations/example/.envrc
@@ -1,22 +1,53 @@
+# region Styling functions
+fmt_var() {
+  printf '\033[38;5;208m%s\033[0m' "$*"
+}
+fmt_val() {
+  printf '\033[1;33m%s\033[0m' "$*"
+}
+log_info() {
+  printf '\033[1;34m%s\033[0m\n' "$*"
+}
+log_progress() {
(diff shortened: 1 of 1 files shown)


soperator/installations/example/.envrc
```diff
@@ -1,35 +1,71 @@
+# region Styling functions
+fmt_var() {
+  printf '\033[38;5;208m%s\033[0m' "$*"
```
theyoprst, line 3: Will github actions show these colors?

theyoprst: It works!
https://github.com/nebius/soperator/actions/runs/28396442651/job/84136302516

theyoprst (approved):

soperator/installations/example/.envrc
dstaroff: I think so. GH actions support ANSI colors and emoji, AFAIK

asteny (approved):

PR #1046, opened 2026-06-30
SCHED-1456: Fix vmagent remotewrite in umbrella chart

Uburro:

### DIFF
--- a/soperator/modules/slurm/locals.tf
+++ b/soperator/modules/slurm/locals.tf
@@ -47,6 +47,8 @@ locals {
 
   active_checks_on_worker_nodes = local.gb300_enabled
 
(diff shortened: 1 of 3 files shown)


soperator/modules/slurm/locals.tf
```diff
@@ -47,6 +47,8 @@ locals {
 
   active_checks_on_worker_nodes = local.gb300_enabled
 
+  public_o11y_tsa_token_writer_source = var.region == "eu-west2" ? "imds" : "file"
```
ali-sattari, line 50: IMDS is available in all regions now. File is not available in new regions. Why not switch it to IMDS by default everywhere and keep file as a fallback?

PR #1047, opened 2026-06-30
SCHED-1913: Batch deletion of leftover disks during cleanup

dstaroff:
## Release Notes

Sped up cluster destruction by replacing sequential disk deletion with parallel execution.


soperator/modules/cleanup/scripts/disk_cleanup.py
```diff
@@ -0,0 +1,496 @@
...
+    print(
+        f"Starting deletion of leftover disk {disk.id} "
+        f"({disk.namespace}/{disk.name})",
+        flush=True,
+    )
+    operation_id = await start_delete_operation(disk, config, start_limiter)
+    if operation_id is None:
+        return await recheck_after_not_found(disk, config)
+
+    print(f"Waiting for disk {disk.id} delete operation {operation_id}", flush=True)
```
asteny: logging?

dstaroff: I see no reason to use it for the script running in `local-exec`

asteny: What do you mean? There will no output with logging?

soperator/modules/cleanup/scripts/disk_cleanup.py
```diff
@@ -0,0 +1,496 @@
...
+    if not exists:
+        print(f"Disk {disk.id} is already absent; task complete", flush=True)
+        return TaskResult(TaskStatus.COMPLETE)
+
+    print(
+        f"Starting deletion of leftover disk {disk.id} "
+        f"({disk.namespace}/{disk.name})",
+        flush=True,
+    )
+    operation_id = await start_delete_operation(disk, config, start_limiter)
```
theyoprst: My initial suggestion about --async usage was incorrect: if we call --async and then call `operation wait`, it's the same as calling it without async once. It can be simplified by removing one call; it's up to you.

soperator/modules/cleanup/scripts/disk_cleanup.py
```diff
@@ -0,0 +1,496 @@
...
+
+
+async def run_delete_task(
+    task: DeleteTask,
+    config: Config,
+    start_limiter: StartRateLimiter,
+) -> TaskResult:
+    disk = task.disk
+
+    exists = await disk_exists(disk, config)
```
theyoprst: Do we really need this call? In the success path, the disk should exist; the delete operation returns an error, and the existence will be rechecked in recheck_after_not_found in any case, right? So it can be removed from there, and the number of API calls will be decreased?

soperator/modules/cleanup/scripts/disk_cleanup.py
```diff
@@ -0,0 +1,496 @@
...
+    "soperator",
+    "nfs-system",
+)
+
+NOT_FOUND_PATTERNS: tuple[str, ...] = (
+    "ResourceNotFound",
+    "disk not found",
+)
+
+DEFAULT_PARALLELISM: Final[int] = 200
```
theyoprst: 200 here, 500 in .envrc. Let's align them? I don't mind which one; maybe 200 is safer. 
(Also, probably remove it from .envrc now at all? If we need to change it, we can always modify the script?)

soperator/modules/cleanup/scripts/disk_cleanup.py
```diff
@@ -0,0 +1,496 @@
...
+        return f"{self.stdout}\n{self.stderr}"
+
+
+class TaskStatus(StrEnum):
+    COMPLETE = "complete"
+    REQUEUE = "requeue"
+
+
+@dataclass(frozen=True)
+class Config:
```
theyoprst: Only parent_id is configurable; the others are constants (if we get rid of parallelism in .envrc).
So maybe we can get rid of the Config class altogether, use DEFAULT_ constants directly, and pass parent_id instead of config where it is needed?
Up to you though

soperator/modules/cleanup/scripts/disk_cleanup.py
dstaroff: What advantages doe this give?

```diff
--- a/soperator/modules/cleanup/scripts/disk_cleanup.py
+++ b/soperator/modules/cleanup/scripts/disk_cleanup.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 import asyncio
 import json
+import logging
 import os
 import random
 import sys
@@ -29,12 +30,23 @@ DEFAULT_INITIAL_REQUEUE_BACKOFF_SECONDS: Final[float] = 1.0
 DEFAULT_MAX_REQUEUE_BACKOFF_SECONDS: Final[float] = 60.0
 DEFAULT_START_RATE_PER_SECOND: Final[float] = 100.0
 DEFAULT_OPERATION_WAIT_TIMEOUT: Final[str] = "5m"
+logger = logging.getLogger("disk_cleanup")
 
 
 class CleanupError(Exception):
     pass
 
 
+def configure_logging() -> None:
+    handler = logging.StreamHandler(sys.stdout)
+    handler.setFormatter(logging.Formatter("%(message)s"))
+
+    logger.handlers.clear()
+    logger.addHandler(handler)
+    logger.setLevel(logging.INFO)
+    logger.propagate = False
+
+
 class CommandError(CleanupError):
     def __init__(self, command: list[str], returncode: int, stdout: str, stderr: str):
         super().__init__(format_command_error(command, returncode, stdout, stderr))
@@ -268,10 +280,11 @@ async def list_disks(config: Config) -> list[Disk]:
                 result.append(disk)
 
         page_token = payload.get("next_page_token") or ""
-        print(
-            f"Scanned disk page {page_number}: {len(items)} items, "
-            f"{len(result)} cleanup candidates so far",
-            flush=True,
+        logger.info(
+            "Scanned disk page %s: %s items, %s cleanup candidates so far",
+            page_number,
+            len(items),
+            len(result),
         )
         if not page_token:
             return result
@@ -360,7 +373,7 @@ async def wait_delete_operation(
 async def recheck_after_not_found(disk: Disk, config: Config) -> TaskResult:
     exists = await disk_exists(disk, config)
     if not exists:
-        print(f"Disk {disk.id} is absent; task complete", flush=True)
+        logger.info("Disk %s is absent; task complete", disk.id)
         return TaskResult(TaskStatus.COMPLETE)
     return TaskResult(TaskStatus.REQUEUE, "disk still exists after not-found response")
 
@@ -374,24 +387,25 @@ async def run_delete_task(
 
     exists = await disk_exists(disk, config)
     if not exists:
-        print(f"Disk {disk.id} is already absent; task complete", flush=True)
+        logger.info("Disk %s is already absent; task complete", disk.id)
         return TaskResult(TaskStatus.COMPLETE)
 
-    print(
-        f"Starting deletion of leftover disk {disk.id} "
-        f"({disk.namespace}/{disk.name})",
-        flush=True,
+    logger.info(
+        "Starting deletion of leftover disk %s (%s/%s)",
+        disk.id,
+        disk.namespace,
+        disk.name,
     )
     operation_id = await start_delete_operation(disk, config, start_limiter)
     if operation_id is None:
         return await recheck_after_not_found(disk, config)
 
-    print(f"Waiting for disk {disk.id} delete operation {operation_id}", flush=True)
+    logger.info("Waiting for disk %s delete operation %s", disk.id, operation_id)
     operation_wait_completed = await wait_delete_operation(operation_id, config)
     if not operation_wait_completed:
         return await recheck_after_not_found(disk, config)
 
-    print(f"Deleted leftover disk {disk.id}", flush=True)
+    logger.info("Deleted leftover disk %s", disk.id)
     return TaskResult(TaskStatus.COMPLETE)
 
 
@@ -413,18 +427,23 @@ async def worker(
         if result.status == TaskStatus.REQUEUE:
             if task.requeues >= config.max_requeue:
                 failures[disk.id] = result.reason
-                print(
-                    f"{name}: disk {disk.id} exceeded max requeue "
-                    f"({config.max_requeue}): {result.reason}",
-                    file=sys.stderr,
-                    flush=True,
+                logger.error(
+                    "%s: disk %s exceeded max requeue (%s): %s",
+                    name,
+                    disk.id,
+                    config.max_requeue,
+                    result.reason,
                 )
             else:
                 delay = requeue_delay(config, task.requeues)
-                print(
-                    f"{name}: requeue disk {disk.id} in {delay:.1f}s "
-                    f"({task.requeues + 1}/{config.max_requeue}): {result.reason}",
-                    flush=True,
+                logger.warning(
+                    "%s: requeue disk %s in %.1fs (%s/%s): %s",
+                    name,
+                    disk.id,
+                    delay,
+                    task.requeues + 1,
+                    config.max_requeue,
+                    result.reason,
                 )
                 await asyncio.sleep(delay)
                 queue.put_nowait(DeleteTask(disk=disk, requeues=task.requeues + 1))
@@ -453,42 +472,43 @@ async def cleanup_disks(config: Config, disks: list[Disk]) -> int:
     await asyncio.gather(*workers, return_exceptions=True)
 
     if failures:
-        print("Failed disk deletions:", file=sys.stderr)
+        logger.error("Failed disk deletions:")
         for disk_id, error in sorted(failures.items()):
-            print(f"- {disk_id}: {error}", file=sys.stderr)
+            logger.error("- %s: %s", disk_id, error)
         return 1
     return 0
 
 
 async def async_main() -> int:
     config = load_config()
-    print(
-        "Disk cleanup config: "
-        f"parallelism={config.parallelism}, "
-        f"page_size={config.page_size}, "
-        f"cli_retries={config.cli_retries}, "
-        f"max_requeue={config.max_requeue}, "
-        f"start_rate_per_second={config.start_rate_per_second}",
-        flush=True,
+    logger.info(
+        "Disk cleanup config: parallelism=%s, page_size=%s, cli_retries=%s, "
+        "max_requeue=%s, start_rate_per_second=%s",
+        config.parallelism,
+        config.page_size,
+        config.cli_retries,
+        config.max_requeue,
+        config.start_rate_per_second,
     )
 
     disks = await list_disks(config)
     if not disks:
-        print("No leftover disks to delete", flush=True)
+        logger.info("No leftover disks to delete")
         return 0
 
-    print(f"Found {len(disks)} leftover disks to delete", flush=True)
+    logger.info("Found %s leftover disks to delete", len(disks))
     return await cleanup_disks(config, disks)
 
 
 def main() -> int:
+    configure_logging()
     try:
         return asyncio.run(async_main())
     except CleanupError as error:
-        print(str(error), file=sys.stderr)
+        logger.error("%s", error)
         return 1
     except KeyboardInterrupt:
-        print("Interrupted", file=sys.stderr)
+        logger.error("Interrupted")
         return 130
```

soperator/modules/cleanup/scripts/disk_cleanup.py
rdjjke: There is a rate limit on the compute side, and it's 100 if I'm not mistaken.
What about asking them?

soperator/modules/cleanup/scripts/disk_cleanup.py
dstaroff: The reason is that the task could be requeued. It can happen if:
1. It failed to check for the disk existence before deleting it
2. It failed to start the deletion operation
3. It failed to wait for the operation to complete
4. It failed to check for the disk existence after deleting it

Nebius CLI fails with `return code != 0` if the disk doesn't exist, but it prints `ResourceNotFound` service message into `stderr`. This is covered by the `NOT_FOUND_PATTERNS`. If it fails with other reason, we retry the whole task few times to be sure we're not hitting the rate limiter or facing network/API flaps. If the task failed on steps 2-3, we ensure the disk still exists before trying to delete it again. If the problem is auth or something else, the disk won't be deleted and will be reported as not deleted at the end of the execution under `Failed disk deletions`.

soperator/modules/cleanup/scripts/disk_cleanup.py
dstaroff: Found it. 100 would be a good default.

soperator/modules/cleanup/scripts/disk_cleanup.py
asteny: Same as using typing here, but newer run mypy :)
I dont thing, that you will use println in Go for logging, so this is a same

soperator/modules/cleanup/scripts/disk_cleanup.py
dstaroff: Due to the changes, this pre-delete get is no longer needed.

With the current flow, `delete --async` is already the source of truth for starting deletion. If the disk is already gone, Nebius CLI returns a non-zero exit with `ResourceNotFound` in `stderr`, and we handle that through `NOT_FOUND_PATTERNS` + `recheck_after_not_found()`. The same recheck path is used after operation wait returns `NOT_FOUND`.

So the pre-delete existence check only adds one API call per disk on the normal path, without improving correctness. I removed it and kept the post-`NOT_FOUND` recheck.

soperator/modules/cleanup/scripts/disk_cleanup.py
dstaroff: Used `logging`

soperator/modules/cleanup/scripts/disk_cleanup.py
dstaroff: I've created dedicated tasks for starting the deletion and waiting for the operation to complete.
The reason is that the rate limiter is only set for the requests (i.e. `disk delete -async`), while the amount of operation in the background is not limited. So we can start deleting disks within `START_DELETE_PARALLELISM` ranges to not hit the rate limiter, and we wait for all disks to be deleted with particular tasks.
This gave a huge boost for the whole `local-exec` time.

soperator/modules/cleanup/scripts/disk_cleanup.py
dstaroff: Set the `START_DELETE_PARALLELISM` to 100.

soperator/modules/cleanup/scripts/disk_cleanup.py
dstaroff: Got rid of `Config` and used constants instead.

asteny (approved):

theyoprst (approved):

PR #1048, opened 2026-07-01
SCHED-1904 Remove extensive checks

asteny:

### DIFF
--- a/soperator/modules/slurm/templates/helm_values/terraform_fluxcd_values.yaml.tftpl
+++ b/soperator/modules/slurm/templates/helm_values/terraform_fluxcd_values.yaml.tftpl
@@ -912,7 +912,6 @@ resources:
               checks:
                 manager:
                   enabledNodeReplacement: ${soperator_checks_node_replacement_enabled}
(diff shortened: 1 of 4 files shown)


ChessProfessor (approved):

soperator/modules/slurm/templates/helm_values/terraform_fluxcd_values.yaml.tftpl
```diff
@@ -912,7 +912,6 @@ resources:
               checks:
                 manager:
                   enabledNodeReplacement: ${soperator_checks_node_replacement_enabled}
-                  enableExtensiveCheck: ${soperator_checks_extensive_check_enabled}
```
itechdima, line 915: do we still have it in Soperator repo?
maybe we can set here hardcoded `false`?
Just in case someone changes the default value in the chart to `true`

asteny: No, we completely remove extensive check https://github.com/nebius/soperator/pull/2671

PR #1058, opened 2026-07-03
More effective resource usage

dstaroff:
## Release Notes

- Decreased default controller VM preset
- Decreased default worker boot disk size
- Switched `controller-spool` disks to `IO_M3` to reduce latency
- Excluded configuration of `etcd` cluster size to auto-scale it on MK8s side to speed up cluster creation

### DIFF
--- a/soperator/installations/example/terraform.tfvars
+++ b/soperator/installations/example/terraform.tfvars
@@ -357,7 +357,7 @@ slurm_nodeset_workers = [
     }
     boot_disk = {
       type                 = "NETWORK_SSD"
-      size_gibibytes       = 512
+      size_gibibytes       = 128
       block_size_kibibytes = 4
     }
     gpu_cluster = {
--- a/soperator/installations/example/terraform.tfvars
+++ b/soperator/installations/example/terraform.tfvars
@@ -316,7 +316,7 @@ slurm_nodeset_controller = {
   size = 1
   resource = {
     platform = "cpu-d3"
-    preset   = "64vcpu-256gb"
+    preset   = "16vcpu-64gb"
   }
   boot_disk = {
     type                 = "NETWORK_SSD"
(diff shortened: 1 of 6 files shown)


soperator/installations/example/terraform.tfvars
```diff
@@ -316,7 +316,7 @@ slurm_nodeset_controller = {
   size = 1
   resource = {
     platform = "cpu-d3"
-    preset   = "64vcpu-256gb"
+    preset   = "16vcpu-64gb"
```
asteny, line 319: But we raise this preset for big clusters in previous task)))

soperator/installations/example/variables.tf
```diff
@@ -513,17 +513,6 @@ variable "k8s_cluster_node_ssh_access_public_ip" {
   default     = false
 }
 
-variable "etcd_cluster_size" {
```
asteny, line 516: Why we remove it? I am using for development with value 1

soperator/installations/example/terraform.tfvars
dstaroff: `SCHED-1896`

soperator/installations/example/variables.tf
dstaroff: `SCHED-1891`

dstaroff: The MK8s has 1 as default and does up-scale as soon as the cluster gets at least one GPU node connected. This allows to deploy the cluster faster at the beginning and have it scaled while the node groups are getting up.
Also, a hand-tuned knob could be wrong at scale.

soperator/installations/example/terraform.tfvars
dstaroff: The controller node showed very low cpu/memory usage (with a high volume of context switches) on the `64vcpu` preset. Dropping to `16vcpu` didn't make anything worse and slightly improved total RPC time and latency.

asteny (approved):

PR #1059, opened 2026-07-03
SCHED-1951: Optional Docker on clusters

Uburro:

### DIFF
--- a/soperator/modules/slurm/main.tf
+++ b/soperator/modules/slurm/main.tf
@@ -80,6 +80,16 @@ resource "helm_release" "soperator_fluxcd_cm" {
 
     dcgm_job_mapping_enabled       = var.dcgm_job_mapping_enabled
     enroot_direct_squashfs_enabled = var.enroot_direct_squashfs_enabled
+    # Cluster-wide toggle for the [program:dockerd] block in the shared jail
+    # supervisord config (customConfigmaps), which is one configmap for the whole
+    # cluster and cannot be made per-nodeset. It is NOT the per-nodeset Docker
+    # switch: each nodeset's Docker is gated separately by its own
+    # node_local_image_storage.enabled in the nodesets values (docker-proxy sidecar
+    # + masking docker/dockerd binaries in the jail). Here we only decide whether to
+    # keep dockerd in the shared supervisord at all: keep it if any nodeset has
+    # image-storage disks (Docker requires them for OCI data), drop it entirely only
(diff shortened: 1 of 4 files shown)


soperator/modules/slurm/main.tf
```diff
@@ -80,6 +80,10 @@ resource "helm_release" "soperator_fluxcd_cm" {
 
     dcgm_job_mapping_enabled       = var.dcgm_job_mapping_enabled
     enroot_direct_squashfs_enabled = var.enroot_direct_squashfs_enabled
+    # Docker needs node-local image-storage disks for storing OCI data.
+    # The jail supervisord config is shared by all workers, so dockerd is kept
+    # if at least one nodeset has image-storage disks enabled.
+    docker_enabled = anytrue([for nodeset in var.worker_nodesets : nodeset.node_local_image_storage.enabled])
```
rdjjke, line 92: Can we also edit the comment for image_storage disks in terraform.tfvars?
To make it clear that they are required for Docker to work?

Uburro: # Cluster-wide toggle for the [program:dockerd] block in the shared jail
    # supervisord config (customConfigmaps), which is one configmap for the whole
    # cluster and cannot be made per-nodeset. It is NOT the per-nodeset Docker
    # switch: each nodeset's Docker is gated separately by its own
    # node_local_image_storage.enabled in the nodesets values (docker-proxy sidecar
    # + masking docker/dockerd binaries in the jail). Here we only decide whether to
    # keep dockerd in the shared supervisord at all: keep it if any nodeset has
    # image-storage disks (Docker requires them for OCI data), drop it entirely only
    # when no nodeset does, so disk-less clusters don't restart-loop a stub dockerd.
    
    it's ok ?

rdjjke: Sorry, I meant adding a comment for `node_local_image_disk` TF variable in `example/terraform.tfvars`.
Something like that:
```
    # Whether to create node-local disks for storing images and container filesystems on each worker node, which are required for Docker container runtime to work.
    # If disabled, only Enroot containers will work.
    # NOTE: `size` must be divisible by 93Gi - https://docs.nebius.com/compute/storage/types#disks-types.
    # ---
    node_local_image_disk = {
      enabled = false
    }
```

Uburro: done

rdjjke (approved):

PR #1073, opened 2026-07-09
SCHED-1893: Scale system component resources with cluster size

theyoprst:
Considered together with nebius/soperator#2735: the slurm-cluster chart dropped `exporterContainer.resources` until that PR, so the exporter tier row only takes effect with it.

## Problem

System/observability component resources are hardcoded for big clusters. Small clusters waste money on oversized CPU nodes and pods, while some components (SPO controller, kruise) OOM on really big ones.

## Solution

- New `modules/sizing_tier`: derives an XS..XL tier from the total worker count and returns per-component pod resources plus CPU node presets for that tier. `sizing_tier_override` forces a tier.
- `modules/slurm` sizes exporter, REST, mariadb, node_configurator, slurm_operator, slurm_checks, kruise (manager + daemon), dcgm_exporter, nfs_server, SPO and the VM/collector stack from the tier. SPO and kruise-manager resources in the fluxcd values are templated instead of hardcoded.
- The example installation defaults controller/accounting/nfs/system node presets from the tier when `resource.preset` is omitted; an explicit preset wins.
- `system_resources` and the `resources_vm_*`/collector variables are replaced by one `component_overrides` object; an entry replaces that component's tier value wholesale. `modules/slurm` input `resources` is renamed to `node_capacity` to separate node capacity from pod sizing.
- Each component row documents which nodes its pods run on, and a capacity guard verifies the presets fit those nodes in every tier (allocatable capacity from `modules/available_resources`; mariadb must leave the accounting node room for slurmdbd). A broken table fails any plan with a message naming the component, the numbers and the node.

## Testing

`terraform test` for `modules/sizing_tier` (17 cases: tier boundaries, overrides, legacy-default and big-cluster parity, capacity invariants) passes; `make terraform-check` now runs module tests, so they also run in CI. Example installation plan checked manually.

## Release Notes

Feature: system component resources and CPU node presets now scale with cluster size via an auto-derived sizing tier; override with `sizing_tier_override` or per-component `component_overrides`. Breaking (Terraform): `system_resources` and `resources_vm_*` variables are replaced by `component_overrides`.

### DIFF
--- a/soperator/installations/example/terraform.tfvars
+++ b/soperator/installations/example/terraform.tfvars
@@ -245,7 +245,7 @@ slurm_nodeset_system = {
   max_size = 24
   resource = {
     platform = "cpu-d3"
-    preset   = "32vcpu-128gb"
+    # preset omitted -> driven by sizing_tier. Set a preset to override.
   }
   boot_disk = {
     type                 = "NETWORK_SSD"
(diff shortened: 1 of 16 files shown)


dstaroff:
LGTM

soperator/installations/example/terraform.tfvars
```diff
@@ -254,69 +254,28 @@ slurm_nodeset_system = {
...
-  }
-  dcgm_exporter = {
-    cpu_cores        = 0.05
-    memory_gibibytes = 0.5
-  }
-}
+# Sizing tier override. The sizing tier is a single knob that scales all system/observability
+# component resources (kruise, VM stack, SPO, collectors, REST, mariadb, ...) and CPU node
+# presets by cluster size. null (default) auto-derives the tier from the worker node count;
+# set "XXS".."XXL" to force it (e.g. "XXS" for e2e).
```
ali-sattari: I see no `XXS` in size module. Is this intentional or a leftover?

theyoprst: Thanks, updated the comment.

theyoprst: Initially, there were also XXS and XXL, but then I got rid of them, we can introduce something special for E2E tests / dev clusters later if needed

dstaroff (approved):

PR #1083, opened 2026-07-16
SCHED-1898: Provide a knob for tuning max in-flight collections per Slurm Exporter's sub-collector

dstaroff:
## Release Notes

Max in-flight collections per Slurm Exporter's sub-collector can now be tuned to enable overlapping collections.

### DIFF
--- a/soperator/installations/example/terraform.tfvars
+++ b/soperator/installations/example/terraform.tfvars
@@ -506,6 +506,9 @@ slurm_exporter_enabled = true
 # Maximum number of concurrent collections per collector in Slurm exporter.
 # By default, 1.
 # ---
+# WARNING: Increasing this value may cause OOM issues on the REST component.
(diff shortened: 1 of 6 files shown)


rdjjke (approved):

soperator/installations/example/terraform.tfvars
```diff
@@ -503,6 +503,11 @@ slurm_login_ssh_root_public_keys = [
 # ---
 slurm_exporter_enabled = true
 
+# Maximum number of concurrent collections per collector in Slurm exporter.
```
theyoprst, line 506: Maybe add a warning that there is a risk of OOM on the rest component, and it is recommended to increase rest nodes resources if you increase this value?

dstaroff: Makes sense. It's stated in PR to Soperator, but not here

theyoprst (approved):

PR #1089, opened 2026-07-17
[CHORE] Disable loki by default

roman-iurkov:
Now that the observability service is live, Loki appears to be redundant and could be disabled by default.


aaroniscode (changes requested):
Hi Roman -- Can you update the comment to point to the new Nebius AI Cloud Observability Logs documentation so anyone using the Terraform will understand that there is a built in replacement for Loki and they don't turn it on and have 2 logging solutions.

roman-iurkov:
> Hi Roman -- Can you update the comment to point to the new Nebius AI Cloud Observability Logs documentation so anyone using the Terraform will understand that there is a built in replacement for Loki and they don't turn it on and have 2 logging solutions.

@aaroniscode added the deprication notice comment to the tfvars file

aaroniscode:
Thanks, Roman — this is heading in the right direction. Could we make one small adjustment? The current link points to the general agent documentation, while I’d like users to land directly on the Observability Logs ingestion guide. Also, “Loki is deprecated” sounds broader than intended; this is specifically about the Loki deployment in this solution.

Suggested replacement:

```hcl
# Loki is disabled by default because the Nebius Observability Agent for Kubernetes
# collects workload logs and forwards them to Observability Logs.
# The agent is enabled by default via `enable_nebius_o11y_agent`.
# Enable `loki.enabled` only if you intentionally want a separate local Loki deployment.
# https://docs.nebius.com/observability/logs/ingest/nebius-o11y-agent
```

This should make the replacement and the risk of enabling two logging paths clearer. Thanks!

PR #1091, opened 2026-07-17
k8s-training: make node group roll-out strategy and capacity reservation configurable

aydinmirac:
## Purpose

The mk8s node group roll-out default is migrating (Q3 2026) from surge-based "create-first" (`max_surge 1 / max_unavailable 0`) to "delete-first" (`max_surge 0 / max_unavailable 1`). Today `k8s-training` sets no roll-out strategy and no reservation policy on its node groups, so both are silently governed by the provider default, meaning the next apply after that migration would change roll-out behavior on existing clusters without any code change.

This is especially impactful for reserved-capacity (capacity block group) customers: a surge node above the reserved count spills into pay-as-you-go billing (`policy = AUTO`) or stalls the roll-out (`policy = STRICT`), and once placed in PAYG it does not migrate back into the block without recreation.

This PR makes both behaviors explicit and configurable so operators can choose the right trade-off for their capacity posture, instead of inheriting a shifting default.

## What changed

**`variables.tf`** — two new optional variables:
- `node_group_strategy` — roll-out pacing on template change: `max_surge` / `max_unavailable` (each `count` or `percent`) and `drain_timeout`.
- `node_group_reservation_policy` — capacity placement: `policy` (`AUTO` | `STRICT` | `FORBID`) and `reservation_ids`, with a validation block restricting `policy` to the allowed values.

Both default to `null`, so unset = inherit the provider default → **fully backward compatible** (no change for existing users who don't set them).

**`main.tf`** — wired the variables into both node groups:
- `strategy = var.node_group_strategy` at the resource level (sibling of `template`) on `cpu-only` and `gpu`.
- `reservation_policy = var.node_group_reservation_policy` inside `template` on `cpu-only` and `gpu`.

**`terraform.tfvars`** — example values documenting the recommended baseline (delete-first + reservation) and the urgent-scale (`AUTO`) alternative.

**`README.md`** — new "Node group roll-out strategy and capacity reservation" section with reference tables for both variables.

## Behavior notes

- `strategy` is resource-level → changing it updates the rule in place with no node roll-out.
- `reservation_policy` is inside `template` → changing it triggers a node roll-out (nodes are re-created to adopt the new placement).

## Testing

Validated on a live CPU-only cluster (`gpu_node_groups = 0`, `enable_gpu_cluster = false`), triggering roll-outs via a `template` change (`cpu_disk_size` / `cpu_nodes_preset`):

- **delete-first** (`max_surge 0 / max_unavailable 1`): nodes replaced strictly one at a time, count dipping 3 → 2 → 3 per node.
- **create-first** (`max_surge 1 / max_unavailable 0`): replacement node joins `Ready` before any old node drains, count 3 → 4 → 3; capacity never dropped below desired.
- `terraform validate` passes; `node_group_reservation_policy` validation rejects values outside `AUTO | STRICT | FORBID`.


roman-iurkov (approved):
LGTM

k8s-training/terraform.tfvars
```diff
@@ -61,7 +84,7 @@ enable_grafana           = true # Enable or disable Grafana® solution by Nebius
 # Local Observability installation
 enable_prometheus = false # Enable or disable Prometheus and Grafana deployment with true or false
 loki = {
-  enabled            = true # Enable or disable Loki deployment with true or false
+  enabled            = false # Enable or disable Loki deployment with true or false
```
aaroniscode: This is unrelated to the PR. Can you remove?

k8s-training/main.tf
```diff
@@ -97,6 +98,7 @@ resource "nebius_mk8s_v1_node_group" "cpu-only" {
       }
     ] : null
     underlay_required = false
+    reservation_policy = var.node_group_reservation_policy
```
aaroniscode: Reservation policy only applies to GPU. I think we remove this for CPU.

k8s-training/variables.tf
```diff
@@ -109,6 +109,61 @@ variable "ssh_public_key" {
...
+  default = null
+
+  validation {
+    condition = var.node_group_reservation_policy == null || contains(
+      ["AUTO", "STRICT", "FORBID"], coalesce(var.node_group_reservation_policy.policy, "AUTO")
+    )
+    error_message = "policy must be one of: AUTO, STRICT, FORBID."
+  }
+
+  # STRICT must name at least one capacity block group.
```
aaroniscode: Can you double-check this. I think you can use STRICT without a CBG.

aaroniscode (changes requested):
looks good with a few review comments requesting changes.

k8s-training/terraform.tfvars
aydinmirac: Reverted back to "true" to keep the previous state

k8s-training/main.tf
aydinmirac: Removed from CPU resource

k8s-training/variables.tf
aydinmirac: Right. [Docs](https://docs.nebius.com/terraform-provider/reference/resources/mk8s_v1_node_group#nested-schema-for-template-reservation_policy) says that "STRICT" will try to launch the instance in Capacity Blocks from `reservation_ids` if provided. I removed the validation of "STRICT"

PR #1095, opened 2026-07-24
chore(soperator): standardize provider declarations

AdamSabry1233:
## What

Adds back the soperator-scoped provider-declaration standardization that was descoped from #1023 (renames) and #1024 (version constraints + tflint enforcement), so soperator can be reviewed/merged independently alongside #1013.

Scoped **strictly to `soperator/`** — no non-soperator files (those remain in #1023/#1024).

### Changes
1. **Rename** the 9 `terraform.tf` provider-declaration files → `provider.tf` (`installations/example/` + `modules/{available_resources,filestore,backups,backups_store,k8s,login,slurm,fluxcd}/`).
2. **Add version constraints** to the `required_providers` entries that only declared a `source`:
   - `nebius     >= 0.6.13, < 0.7.0`
   - `kubernetes >= 3.0.1,  < 4.0.0`
   - `units      >= 1.1.1,  < 2.0.0`
   - `local      >= 2.5.3,  < 3.0.0`
3. **Enable enforcement**: flip `version = false` → `version = true` in the 6 soperator `.tflint.hcl` files (example, available_resources, filestore, k8s, login, slurm). `modules/labels/.tflint.hcl` has no `terraform_required_providers` rule, so it is untouched.

## Root cause of the previously-failing "Soperator Terraform Checks"

The CI job runs `make terraform-check` (fmt + `terraform init -upgrade` / `validate` on every module and installation + module tests).

The earlier failure was **not** tflint — it was the **nebius version constraint**. soperator's nebius `source` is the **public `nebius/nebius`** registry, which only serves **0.6.x**. The originally-proposed `nebius >= 0.5.196, < 0.6.0` (proven only against the *private* `terraform-provider.storage.eu-north1.nebius.cloud/nebius/nebius` source used by non-soperator dirs) produces:

```
Error: no available releases match the given constraints >= 0.5.196, < 0.6.0
```

→ `terraform init` fails → job fails.

**Fix:** use a 0.6.x range that resolves on the public source and matches #1013's direction (`>= 0.6.13`), with an upper bound consistent with the repo's constraint style: **`nebius >= 0.6.13, < 0.7.0`** (resolves to 0.6.32).

`make terraform-check` now passes locally: fmt clean, all modules + `installations/example` validate, and 22/22 `sizing_tier` module tests pass.


soperator/modules/backups_store/provider.tf
```diff
@@ -0,0 +1,8 @@
+terraform {
+  required_providers {
+    nebius = {
+      source  = "nebius/nebius"
+      version = ">= 0.6.13, < 0.7.0"
```
theyoprst, line 5: Better to allow incrementing the minor version: `>= 0.6.13, < 1.0.0`.
nebius/nebius doesn't break backward compatibility, and there were no major problems with the latest migration from 0.5 to 0.6; Terraform was compatible. `< 1.0.0` could be omitted too, but if version 1.0.0 is ever released, it will probably be something completely new, so it's probably a good idea to have it.

soperator/installations/example/provider.tf
theyoprst, line 7: It's not consistent with others now

PR #1099, opened 2026-07-28
Add snapshot-backed boot disks to vm-instance

realvz:

Adds optional snapshot-backed boot disks to the `vm-instance` solution. Existing deployments continue to use the default Ubuntu image unless `boot_disk_snapshot_id` is set.

## What changed

- Added `boot_disk_snapshot_id` to the root solution and instance module.
- Creates each VM boot disk from the selected project-local snapshot.
- Validates that the snapshot belongs to the project, is ready, and fits within the configured disk size.
- Raised the Nebius provider minimum to `0.5.240`, which introduced `source_snapshot_id`.
- Documented snapshot requirements and full-disk clone caveats, including retained guest identity and cloud-init state.

## Validation

- `terraform fmt -check -recursive vm-instance modules/instance`
- `terraform validate -no-color` with Terraform 1.11.4 in Linux

### DIFF
--- a/modules/instance/provider.tf
+++ b/modules/instance/provider.tf
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     nebius = {
       source  = "terraform-provider.storage.eu-north1.nebius.cloud/nebius/nebius"
-      version = ">= 0.5.55"
+      version = ">= 0.5.240"
     }
   }
 }
--- a/vm-instance/provider.tf
+++ b/vm-instance/provider.tf
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     nebius = {
       source  = "terraform-provider.storage.eu-north1.nebius.cloud/nebius/nebius"
-      version = ">= 0.4.24"
+      version = ">= 0.5.240"
     }
   }
 }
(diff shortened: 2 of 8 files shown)


modules/instance/provider.tf
```diff
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     nebius = {
       source  = "terraform-provider.storage.eu-north1.nebius.cloud/nebius/nebius"
```
pbutler, line 4: Let's use the nebius public provide since we need to start doing that.  The equivalent public provider block should look like this but check me (https://docs.nebius.com/terraform-provider/manage/migrate)
```hcl 
{ 
    source = "nebius/nebius"
    version = ">= 0.6.31"
}
```

vm-instance/provider.tf
pbutler, line 1: Let's use the nebius public provide since we need to start doing that.  The equivalent public provider block should look like this but check me (https://docs.nebius.com/terraform-provider/manage/migrate)
```hcl 
{ 
    source = "nebius/nebius"
    version = ">= 0.6.31"
}
```

pbutler (changes requested):
Looks good to me but we should start using the public nebius terraform provider when possible

aaroniscode (changes requested):
Hi @realvz , thanks for the PR!

I agree with @pbutler 's comments to switch to the public provider. The only feedback I have is to add some additional detail to the README in your Example 5 around how `source_snapshot_id` is immutable.

Something like:
> Treat boot_disk_snapshot_id as a creation-time setting. Setting, changing, or removing it for an existing deployment replaces the boot disk and can cause data loss and VM interruption.

PR #1106, opened 2026-07-30
SCHED-1028 Set UCX_NET_DEVICES and NCCL_IB_HCA env var on workers

asteny:

### DIFF
--- a/soperator/modules/slurm/templates/helm_values/flux_release_nodesets.yaml.tftpl
+++ b/soperator/modules/slurm/templates/helm_values/flux_release_nodesets.yaml.tftpl
@@ -172,6 +172,17 @@ nodesets:
                     path: /sys
                     type: Directory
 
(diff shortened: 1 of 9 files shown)


soperator/modules/slurm/templates/helm_values/flux_release_nodesets.yaml.tftpl
```diff
@@ -172,6 +172,17 @@ nodesets:
                     path: /sys
                     type: Directory
 
+              %{~ if nodeset.nccl_network_profile != null ~}
```
rdjjke: In different places, it's "nccl_network_profile", "nccl_network_profiles", or "nccl_network_vars".
What about using a single term? (I personally think that "vars" is better, because "NCCL profile" can be read as something related to NCCL profiling, which is a different thing)

rdjjke (approved):

PR #1113, opened 2026-08-01
Add unified NCCL test runner for MK8s clusters

AdamSabry1233:
## Summary
Adds a single, reusable NCCL benchmarking toolchain for Nebius MK8s clusters that **auto-adapts to the GPU node type** (H100/H200/B200/B300) instead of hardcoding per-hardware assumptions that break when switching clusters.

## What it auto-detects per run
- **Node capacity** (CPU/mem/GPU) → worker resource sizing
- **Active InfiniBand devices** via `ibv_devinfo` (e.g. B300 `mlx5_4..11` vs H200 `mlx5_0..7`)
- **CPU/NUMA topology** → process binding `pe = CPUs/GPUs`. *This fixes the B300→H200 breakage where a hardcoded `pe=24` needs 96 hwthreads/NUMA and aborts on H200's 64/NUMA.*
- **MPIJob API version** (`v1` vs `v2beta1`) → correct manifest variant
- **Available node count** → caps host counts so sweeps never hang on unschedulable jobs

## Behavior
Runs `all_reduce` + `alltoall` over NVLink (single-node) and InfiniBand (multi-node), takes a PID lock so two runners can't stomp each other's pods, retries transient launcher glitches, and auto-generates a markdown `report.md`.

## Files
- `nccl_runner_unified.sh` — orchestrator
- `nccl-test-template.yaml` — parameterized MPIJob
- `generate-report.sh` — raw logs → report
- `README.md`

## Validation
Verified end-to-end on a 2-node H200 cluster: all 4 jobs pass, cross-node runs confirmed `Using network IB` across all 8 HCAs. Peak busbw: all_reduce ~482 GB/s (NVLink & IB), alltoall ~347 (NVLink) / ~92 (IB), zero data errors.

## Scope
Nebius x86 + InfiniBand types (H100/H200/B200/B300). Grace-ARM (GH200/GB200) would need an aarch64 image and re-validated binding math.

### DIFF
--- a/k8s-training/k8s-runner-unified/README.md
+++ b/k8s-training/k8s-runner-unified/README.md
@@ -78,11 +78,16 @@ Regenerate a report from existing logs at any time:
 | `NODE_GROUP_ID` | — | **Set per cluster** |
 | `RESERVE_CPU_CORES` / `RESERVE_MEM_GI` | `4` / `50` | Headroom left for kubelet/daemonsets |
 | `MAX_LAUNCH_ATTEMPTS` | `6` | Retry budget for the transient launcher glitch |
-| `HOSTS` | `(1 2 3 4)` | Host counts to sweep (auto-capped to available nodes) |
+| `HOSTS` | `(1 2 3 4)` | Host counts to sweep. Auto-capped **down** to available Ready nodes, but never scales **up** past the values listed — on clusters larger than 4 nodes, edit this to test at full scale (e.g. `(1 2 4 8 16)` for a 16-node group). |
 | `TESTS` | `(all_reduce alltoall)` | Add `all_gather reduce_scatter reduce gather broadcast scatter` for a fuller sweep |
 
 ## Scope & caveats
 
+- **Sweep size is set by `HOSTS`, not auto-scaled.** The runner caps requested
+  counts down to the available Ready nodes so it never hangs, but it won't test
+  beyond the largest value in `HOSTS`. For clusters bigger than 4 nodes, add
+  higher counts (powers of two up to your node count, plus the full count) so the
+  full-cluster fabric actually gets measured.
 - **Supported as-is:** Nebius **x86 + InfiniBand** GPU types — H100, H200, B200, B300.
 - **Not supported as-is:** Grace-ARM types (GH200 / GB200) — aarch64, so the x86
   image won't run and the binding math should be re-validated.
--- a/k8s-training/k8s-runner-unified/README.md
+++ b/k8s-training/k8s-runner-unified/README.md
@@ -88,6 +88,10 @@ Regenerate a report from existing logs at any time:
   beyond the largest value in `HOSTS`. For clusters bigger than 4 nodes, add
   higher counts (powers of two up to your node count, plus the full count) so the
   full-cluster fabric actually gets measured.
+  - *Example — 100-node cluster:* set `HOSTS=(1 2 4 8 16 32 64 100)`. Powers of
+    two give a clean scaling curve, and `100` measures the full cluster. Leaving
+    the default `(1 2 3 4)` would test only 4 of your 100 nodes and tell you
+    nothing about the fabric at scale.
 - **Supported as-is:** Nebius **x86 + InfiniBand** GPU types — H100, H200, B200, B300.
 - **Not supported as-is:** Grace-ARM types (GH200 / GB200) — aarch64, so the x86
   image won't run and the binding math should be re-validated.
--- a/k8s-training/k8s-runner-unified/README.md
+++ b/k8s-training/k8s-runner-unified/README.md
@@ -29,12 +29,16 @@ runs `all_reduce` + `alltoall` over both NVLink (single-node) and InfiniBand
    hwthreads/NUMA and aborts on H200's 64/NUMA.)*
 4. **MPIJob API version** — `v1` vs `v2beta1` (installs differ per cluster);
    `launcherCreationPolicy` is only emitted on `v2beta1`.
-5. **Host-count cap** — requested host counts are capped to the number of nodes
-   actually in the node group, so the sweep never hangs on unschedulable jobs.
+5. **Host-count cap** — requested host counts are capped to the number of
+   **Ready, schedulable** nodes in the node group (Cordoned / NotReady nodes are
+   excluded), so the sweep never hangs on unschedulable jobs.
 
-It also takes a **PID lock** (`/tmp/nccl_runner_unified.lock`) so two runners can't
-share the namespace and delete each other's pods, and retries a launch up to
-`MAX_LAUNCH_ATTEMPTS` times if it fails to start.
+Every run creates **uniquely-named, labelled resources** (`nccl-test-<pid>-<epoch>`,
+labelled `nccl-runner/run=<id>`), so two engineers can run against the same
+namespace without touching each other's MPIJob or pods. A local **PID lock**
+(`/tmp/nccl_runner_unified.lock`) additionally prevents accidental double-runs on
+the same workstation, and a launch is retried up to `MAX_LAUNCH_ATTEMPTS` times if
+it fails to start.
 
 ## Usage
 
@@ -60,9 +64,10 @@ Regenerate a report from existing logs at any time:
 ./generate-report.sh results/nccl-<timestamp>
 ```
 
-> **One runner per namespace at a time** — the lock enforces this. Two concurrent
-> runners on the same namespace force-delete each other's pods and produce spurious
-> launcher failures.
+> **Concurrent runs are isolated by unique resource names + labels**, so multiple
+> engineers can share a namespace safely. Cleanup is always scoped to a run's own
+> MPIJob and cascaded gracefully — the runner never force-deletes pods or touches
+> pods it doesn't own.
 
 ## Configuration (top of `nccl_runner_unified.sh`)
 
--- a/k8s-training/k8s-runner-unified/README.md
+++ b/k8s-training/k8s-runner-unified/README.md
@@ -1,0 +1,87 @@
+# Unified NCCL Test Runner
+
+A single, reusable NCCL benchmarking toolchain for Nebius MK8s clusters.
+It **auto-adapts to the GPU node type it is pointed at** (H100 / H200 / B200 / B300)
+instead of hardcoding per-hardware assumptions that break when switching clusters,
+runs `all_reduce` + `alltoall` over both NVLink (single-node) and InfiniBand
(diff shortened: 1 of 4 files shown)


k8s-training/k8s-runner-unified/nccl_runner_unified.sh
```diff
@@ -0,0 +1,339 @@
...
+OUTPUT_PATH=results/nccl-$(date +"%Y-%m-%d_%H-%M-%S")
+mkdir -p "$OUTPUT_PATH"
+
+for HOST_NUM in "${HOSTS[@]}"; do
+  for TEST in "${TESTS[@]}"; do
+    echo "=== Starting $TEST on $HOST_NUM host(s) ==="
+
+    # Clean up any leftover release from a prior iteration before starting
+    kubectl delete mpijob nccl-test -n "$NAMESPACE" --ignore-not-found || true
+    kubectl delete pods --all -n "$NAMESPACE" --grace-period=0 --force || true
```
aaroniscode: Could we avoid `--all --grace-period=0 --force` here? `--all` can delete unrelated pods in the namespace, while `--force` deletion removes Pod objects without confirming their processes have stopped. Since these pods are owned by the MPIJob, we should use scoped, graceful cleanup instead:
```
kubectl delete mpijob nccl-test -n "$NAMESPACE" \
  --ignore-not-found --cascade=foreground --wait=true --timeout=60s || true
```
If explicit pod cleanup is still needed, it should target only runner-owned pods by label, with force deletion reserved as a last-resort fallback.

k8s-training/k8s-runner-unified/nccl_runner_unified.sh
```diff
@@ -0,0 +1,339 @@
...
+trap 'rm -f "$LOCKFILE"' EXIT
+
+# ============ CONFIGURE FOR YOUR CLUSTER ============
+NAMESPACE=nccl-tests
+TEMPLATE=nccl-test-template.yaml
+IMAGE="cr.eu-north1.nebius.cloud/e00b94r7bkvywphmn6/nccl-tests:v2.18.3-cudav13.2.1-ncclv2.30.4-1-hpcxv2.26"
+
+# GPU node group ID — get via:
+#   kubectl get nodes -o custom-columns='NAME:.metadata.name,GROUP:.metadata.labels.nebius\.com/node-group-id,GPU:.status.capacity.nvidia\.com/gpu'
+NODE_GROUP_ID="mk8snodegroup-e02j9bmhyza8zwvj9c"
```
aaroniscode: I think we should remove an actual ID and replace with a placeholder

k8s-training/k8s-runner-unified/README.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+Regenerate a report from existing logs at any time:
+
+```bash
+./generate-report.sh results/nccl-<timestamp>
+```
+
+> **One runner per namespace at a time** — the lock enforces this. Two concurrent
+> runners on the same namespace force-delete each other's pods and produce spurious
```
aaroniscode: We shouldn't be force deleting any pods at all, in my opinion.

k8s-training/k8s-runner-unified/nccl_runner_unified.sh
```diff
@@ -0,0 +1,339 @@
...
+if [ -z "$ALL_DEVICES" ]; then
+  echo "ERROR: no IB devices found at all on this node (ibv_devices returned nothing)."
+  kubectl delete pod ib-discovery -n "$NAMESPACE" --ignore-not-found >/dev/null 2>&1
+  exit 1
+fi
+
+ACTIVE_DEVICES=""
+for dev in $ALL_DEVICES; do
+  INFO=$(kubectl exec -n "$NAMESPACE" ib-discovery -- ibv_devinfo -d "$dev" 2>/dev/null)
+  LINK_LAYER=$(echo "$INFO" | grep -oP 'link_layer:\s*\K\S+')
```
aaroniscode: The `-P` flag isn't valid for `grep` on MacOS.

k8s-training/k8s-runner-unified/generate-report.sh
```diff
@@ -0,0 +1,115 @@
...
+fi
+
+REPORT="$RESULTS_DIR/report.md"
+
+# awk program shared by the summary and per-test sections: emits genuine NCCL
+# data rows only (13 fields, field 3 a real dtype), so NCCL_DEBUG noise is dropped.
+DATA_ROWS='NF==13 && $3 ~ /^(float|double|half|bfloat16|int|int8|uint8|char)$/'
+
+# Collect the log files, sorted by host count then test name for a stable order.
+mapfile -t LOGS < <(ls "$RESULTS_DIR"/*.log 2>/dev/null | sort)
```
aaroniscode: `mapfile` is not available by default on MacOS.

k8s-training/k8s-runner-unified/nccl_runner_unified.sh
```diff
@@ -0,0 +1,339 @@
...
+# whatever cluster/GPU type it's pointed at, instead of hardcoding either.
+#
+# Usage: ./nccl-runner-unified.sh
+
+set -e
+
+# Guard against a second concurrent runner. Two runners sharing the nccl-tests
+# namespace delete each other's launcher/worker pods mid-launch, which looks
+# exactly like random launcher crashes ("lost communication") — don't allow it.
+LOCKFILE="/tmp/nccl_runner_unified.lock"
```
aaroniscode, line 15: Two engineers running against the same namespace from different workstations both acquire their local lock, then manipulate the same fixed `nccl-test` resources and interfere with each other. 

Ideally, you would use an atomically acquired Kubernetes Lease/ConfigMap keyed by namespace.

But the simple Kubernetes solution is to use unique per-run resource names and labels.

k8s-training/k8s-runner-unified/nccl_runner_unified.sh
```diff
@@ -0,0 +1,339 @@
...
+# ======================================================
+
+echo "=== Detecting node capacity for node group $NODE_GROUP_ID ==="
+NODE_NAMES=($(kubectl get nodes -l "nebius.com/node-group-id=$NODE_GROUP_ID" -o jsonpath='{.items[*].metadata.name}'))
+NODE_NAME="${NODE_NAMES[0]}"
+if [ -z "$NODE_NAME" ]; then
+  echo "ERROR: no nodes found matching node-group-id=$NODE_GROUP_ID"
+  exit 1
+fi
+NODE_COUNT=${#NODE_NAMES[@]}
```
aaroniscode, line 108: This counts every node carrying the label, including Cordoned and `NotReady` nodes. 

For example, a nodegroup with 4 nodes with one unavailable node still schedules the four-host sweep; with `WaitForWorkersReady`, no launcher appears and `watch_job` waits the full 1500 seconds. 

Filter for Ready nodes with `spec.unschedulable != true`.

k8s-training/k8s-runner-unified/generate-report.sh
```diff
@@ -0,0 +1,115 @@
...
+    if grep -q "Using network IB" "$f" 2>/dev/null; then
+      transport=$([ "$hosts" -gt 1 ] 2>/dev/null && echo "InfiniBand" || echo "NVLink (IB init'd)")
+    else
+      transport="?"
+    fi
+    # peak busbw (field 8) and its size (field 1); errors if any field 9/13 nonzero (N/A ok)
+    read -r peak psize < <(awk "$DATA_ROWS"' {if($8+0>max){max=$8+0;sz=$1}} END{printf "%.1f %s\n", max, sz}' "$f")
+    errs=$(awk "$DATA_ROWS"' (($9!="N/A" && $9+0!=0)||($13!="N/A" && $13+0!=0)){c++} END{print c+0}' "$f")
+    errflag=$([ "$errs" -eq 0 ] && echo "none ✓" || echo "$errs ⚠️")
+    echo "| $test | $hosts | $scope | $transport | ${peak} GB/s | ${psize}B | $errflag |"
```
aaroniscode, line 77: The runner intentionally writes a log after launch failures and timeouts, but these awk expressions return peak 0 and zero error rows when no NCCL data exists. 

Codex reproduced a summary row of `0.0 GB/s | B | none ✓` for a log containing only a launcher failure, while the detail section correctly said no data was found. Check for at least one parsed row before calculating the summary and mark the run incomplete or failed rather than showing a green correctness result.

aaroniscode (changes requested):
Thanks for the PR. I made some comments.

k8s-training/k8s-runner-unified/nccl_runner_unified.sh
AdamSabry1233: Switched to unique per-run names + nccl-runner/run labels; all lookups/cleanup scoped to them. Lock now just guards same-machine double-runs.

k8s-training/k8s-runner-unified/nccl_runner_unified.sh
AdamSabry1233: Replaced with a placeholder + startup validation.

k8s-training/k8s-runner-unified/nccl_runner_unified.sh
AdamSabry1233: Now counts only Ready + schedulable nodes.

k8s-training/k8s-runner-unified/nccl_runner_unified.sh
AdamSabry1233: swapped for portable awk

k8s-training/k8s-runner-unified/nccl_runner_unified.sh
AdamSabry1233: Now scoped --cascade=foreground --wait; dropped --all --force.

k8s-training/k8s-runner-unified/README.md
AdamSabry1233: Removed the behavior and the wording.

k8s-training/k8s-runner-unified/generate-report.sh
AdamSabry1233: Replaced with a portable read loop.

k8s-training/k8s-runner-unified/generate-report.sh
AdamSabry1233: Requires > or equal to 1 data row now; empty logs show X no data.

aaroniscode (approved):
Thanks for making the changes @AdamSabry1233 !  Approved

PR #1114, opened 2026-08-01
Add GPU soak test for MK8s clusters

AdamSabry1233:
## Summary
Adds a Kubernetes-native GPU soak test for MK8s clusters — the MK8s equivalent of the cloudmeter auto-loader: sustained GPU compute + HBM fill + InfiniBand/NVLink stress with structured pass/fail output and a report.

## How it works
Runs as a **PyTorchJob** with a single PyTorch image on every pod. `torchrun` launches one `torch.distributed` rank per GPU, and each rank concurrently:
1. **Fills HBM** with a resident FP16 tensor (default 75% of device memory)
2. **Stresses compute** with a continuous FP16 matmul loop (drives util to ~100%)
3. **Stresses the fabric** with a timed, verified `all_reduce` every iteration

Because PyTorch bundles NCCL, the same container does the memory fill and the collective — no launcher/worker image split — and the collective's exit code is authoritative (a real NCCL failure fails the job).

## Files
- `run-soak-test.sh` — driver
- `scripts/soak.py` — per-rank soak workload
- `scripts/monitor.sh` — monitoring
- `templates/pytorchjob.yaml` — PyTorchJob manifest
- `README.md`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/k8s-training/gpu-soak-test/run-soak-test.sh
+++ b/k8s-training/gpu-soak-test/run-soak-test.sh
@@ -1,0 +1,311 @@
+#!/bin/bash
+# =============================================================================
+# GPU Soak Test Runner (PyTorchJob / torch.distributed)
+# Equivalent to cloudmeter auto-loader for MK8s clusters.
+# Runs sustained GPU compute + HBM fill + NCCL all_reduce (IB/NVLink) stress
+# for a configurable duration, then emits a structured pass/fail report.
+#
+# Usage:
+#   ./run-soak-test.sh [duration_seconds] [gpu_node_count]
+#
+# Examples:
+#   ./run-soak-test.sh              # 1 hour, 2 nodes (defaults)
+#   ./run-soak-test.sh 7200 2       # 2 hours, 2 nodes
+#   ./run-soak-test.sh 3600 4       # 1 hour, 4 nodes
+#
+# Environment overrides:
+#   SOAK_IMAGE          Container image (default is a Hopper/H100 CUDA 12.3 image;
+#                       set a CUDA 12.6+ arm64 image for Blackwell B200/GB200/GB300)
+#   HBM_FILL_FRACTION   Fraction of HBM to fill per GPU (default 0.75)
+#   MAX_TEMP            Overtemp threshold in °C (default 83; raise for Blackwell)
+#   AUTO_CLEANUP=y      Delete the namespace at the end without prompting
+#
+# What it tests (equivalent to auto-loader):
+#   - Sustained GPU utilization across all GPU nodes
+#   - Configurable HBM fill (dynamically sized to GPU type)
+#   - InfiniBand/NVLink stress via repeated NCCL all_reduce (verified + busbw)
+#   - GPU temperature monitoring (fail if > 83°C)
+#   - XID error detection (best-effort; flagged UNVERIFIED if unavailable)
+#   - Node health monitoring
+#   - Structured pass/fail output
+# =============================================================================
+set -euo pipefail
+
+DURATION="${1:-3600}"
+NODE_COUNT="${2:-2}"
+NAMESPACE="gpu-soak"
+SLOTS=8
+HBM_FILL_FRACTION="${HBM_FILL_FRACTION:-0.75}"
+# Container image for the workload. The default is a Hopper-era PyTorch (CUDA
+# 12.3) that works on H100/H200. It will NOT run on Blackwell (B200/GB200/GB300)
+# — those need a newer CUDA 12.6+ image, and Grace-based GB200/GB300 also need an
+# arm64/sbsa image. Override per platform, e.g.:
+#   SOAK_IMAGE=nvcr.io/nvidia/pytorch:<blackwell-arm64-tag> ./run-soak-test.sh ...
+SOAK_IMAGE="${SOAK_IMAGE:-nvcr.io/nvidia/pytorch:24.01-py3}"
+SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)"
+REPORT_FILE="soak-report-$(date +%Y%m%d_%H%M%S).txt"
+START_TIME=$(date -u)
+START_EPOCH=$(date +%s)
+
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+BLUE='\033[0;34m'
+NC='\033[0m'
+
+# Delete the test namespace. Guarded so it runs at most once.
+CLEANED=0
+cleanup_ns() {
+  [ "$CLEANED" = "1" ] && return
+  CLEANED=1
+  echo ""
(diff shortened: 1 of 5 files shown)


k8s-training/gpu-soak-test/run-soak-test.sh
```diff
@@ -0,0 +1,396 @@
...
+
+# Run monitor. set +e so a FAILED soak still produces a report and cleans up
+# (bash monitor.sh returning non-zero must NOT abort this script under set -e).
+set +e
+bash "$SCRIPT_DIR/scripts/monitor.sh"
+MONITOR_EXIT=$?
+set -e
+
+# Collect final master logs (rank 0 prints the summary + BUSBW lines)
+MASTER_LOGS=$(kubectl logs --request-timeout=30s -n "$NAMESPACE" -l training.kubeflow.org/replica-type=master 2>/dev/null | tail -40)
```
pbutler: Here you're expected `kubectl logs` to output all lines of the logs, but this isn't the case when specifying `-l` it only outputs the last 10 lines: https://kubernetes.io/docs/reference/kubectl/generated/kubectl_logs/

k8s-training/gpu-soak-test/run-soak-test.sh
```diff
@@ -0,0 +1,396 @@
...
+END_TIME=$(date -u)
+END_EPOCH=$(date +%s)
+ACTUAL_DURATION=$(( END_EPOCH - START_EPOCH ))
+
+# =============================================================================
+# GENERATE REPORT
+# =============================================================================
+generate_report() {
+  local LOG_FILE
+  LOG_FILE=$(ls -t "$SCRIPT_DIR"/soak-monitor-*.log 2>/dev/null | head -1)
```
pbutler, line 280: `soak-monitor-*.log` is relative to `SCRIPT_DIR` in `monitor.sh:26` looks in CWD, those should be the same as I read it but it's worth a look to verify that they are.

k8s-training/gpu-soak-test/scripts/monitor.sh
```diff
@@ -0,0 +1,290 @@
...
+# result lives in the debugger POD's logs, not inline. We therefore launch the
+# debugger, locate the pod it creates (named node-debugger-<node>-<rand>), wait
+# for it to complete, read the count from its logs, then delete it.
+# Echoes an integer count, or "__FAILED__" if it could not run.
+check_xid_on_node() {
+  local node="$1" dbg="" waited=0 phase="" out=""
+  # `grep -c` exits 1 when the count is zero, which would mark the debugger pod
+  # Failed on a HEALTHY node — so append `|| true` to keep the container clean.
+  kubectl debug node/"$node" --image=ubuntu --profile=sysadmin -q \
+    -- chroot /host sh -c 'dmesg 2>/dev/null | grep -ic xid || true' >/dev/null 2>&1 || true
```
pbutler: dmesg outputs all messages in the ring buffer, so even ones older that might be irrelevant, look into adding `--since` to dmesg.

k8s-training/gpu-soak-test/run-soak-test.sh
```diff
@@ -0,0 +1,396 @@
...
+#   - Structured pass/fail output
+# =============================================================================
+set -euo pipefail
+
+DURATION="${1:-3600}"
+# Exported so monitor.sh can bound its poll loop to the soak duration (+ buffer).
+export SOAK_DURATION_SECONDS="$DURATION"
+NODE_COUNT="${2:-2}"
+NAMESPACE="gpu-soak"
+SLOTS=8
```
pbutler: 4 GPUS / GB node so this will cause problems.

k8s-training/gpu-soak-test/scripts/monitor.sh
```diff
@@ -0,0 +1,290 @@
...
+if [ "$XID_COUNT" -gt "0" ]; then
+  fail "$XID_COUNT XID error(s) detected — potential GPU hardware issue"
+  OVERALL_PASS=false
+elif [ "$XID_UNVERIFIED" -gt "0" ]; then
+  warn "XID check could not run on $XID_UNVERIFIED node(s) — GPUs NOT certified XID-clean"
+else
+  pass "No XID errors"
+fi
+
+if [ "$LOW_UTIL_COUNT" -gt "5" ]; then
```
pbutler: README indicates this should be 10

pbutler (changes requested):
Looks pretty good but I added some comments.  Let me know if you have any questions.

k8s-training/gpu-soak-test/run-soak-test.sh
AdamSabry1233: Good catch — fixed by adding --tail=-1. You're right that -l caps output at the last 10 lines per pod (naming a
 pod directly returns everything), so the | tail -40 was misleading and the summary lines were only landing in 
 the report by luck. --tail=-1 forces the full log. Validated on H200 — busbw/iteration counts now parse reliably.

k8s-training/gpu-soak-test/run-soak-test.sh
AdamSabry1233: Verified — they were not guaranteed to match. monitor.sh wrote soak-monitor-*.log to the current working directory, while the report reads from $SCRIPT_DIR. They only lined up when you happened to run from inside the
folder. Fixed by exporting SOAK_LOG_DIR="$SCRIPT_DIR" and having the monitor write there, so it works regardless of where the script is invoked from.

k8s-training/gpu-soak-test/scripts/monitor.sh
AdamSabry1233: Done — added dmesg --since scoped to the run's start time (computed host-side so it matches the kernel log's clock). Previously it counted every XID line in the ring buffer, including ones from before the soak even started. Now only XIDs during the run window count against it.

k8s-training/gpu-soak-test/run-soak-test.sh
AdamSabry1233: Fixed — SLOTS is now auto-detected from the node's nvidia.com/gpu allocatable (falls back to 8 only if unreadable, and still overridable via SLOTS=<n>). You're right that hardcoding 8 breaks non-8-GPU nodes. Note: the GB300/DRA variant (#1135) already detects this in a DRA-aware way, so I deliberately did not copy this x86 detection there. Validated on H200 — detects 8 correctly.

k8s-training/gpu-soak-test/scripts/monitor.sh
AdamSabry1233: You're right — the README (and the report generator) use 10, but the monitor used 5. Aligned the monitor to -ge 10 so all three agree ("fewer than 10 low-util events = pass").

pbutler (approved):
Changes look good to me.  Thanks!

PR #1115, opened 2026-08-01
Add checkpoint/restore test to filesystem CSI validation

AdamSabry1233:
## Summary
Adds `05-run-checkpoint-restore-test.sh` to the existing `filesystem-csi-validation` suite.

It validates that model checkpoints can be written to and restored from the **Nebius Shared Filesystem** correctly across nodes — testing both **same-node** and **cross-node** restore on **all GPU nodes** in the cluster. This complements the existing CSI smoke (`02`) and RWX cross-node (`03`) tests.

## Files
- `filesystem-csi-validation/05-run-checkpoint-restore-test.sh`

### DIFF
--- a/k8s-training/filesystem-csi-validation/05-run-checkpoint-restore-test.sh
+++ b/k8s-training/filesystem-csi-validation/05-run-checkpoint-restore-test.sh
@@ -45,9 +45,15 @@ source "${SCRIPT_DIR}/common.sh"
 
 CHECKPOINT_SIZE_GB="${CHECKPOINT_SIZE_GB:-4}"
 CHECKPOINT_PATH="/data/checkpoint"
-WRITER_POD="checkpoint-writer"
-SAME_NODE_READER_POD="checkpoint-reader-same"
-PVC_NAME="checkpoint-restore-pvc"
+# Unique per-run suffix so two runs sharing the namespace can't collide on fixed
+# resource names or cross-delete each other's pods/PVC during cleanup.
+RUN_ID="$$-$(date +%s)"
+WRITER_POD="checkpoint-writer-${RUN_ID}"
+SAME_NODE_READER_POD="checkpoint-reader-same-${RUN_ID}"
+PVC_NAME="checkpoint-restore-pvc-${RUN_ID}"
+# part-of matches the sibling scripts, so the shared 04-cleanup also catches these;
+# the run label lets this script tear down exactly its own resources.
+RUN_LABEL="checkpoint-restore/run=${RUN_ID}"
 FAILED=0
 NODES_TESTED=0
 NODES_PASSED=0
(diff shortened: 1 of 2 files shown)


k8s-training/filesystem-csi-validation/05-run-checkpoint-restore-test.sh
```diff
@@ -0,0 +1,483 @@
...
+# DETECT GPU NODES DYNAMICALLY
+# Only test GPU nodes — CPU nodes don't run training jobs so cross-node
+# restore from a CPU node is not a relevant production scenario.
+# =============================================================================
+log_step "Detecting GPU nodes in cluster"
+
+# Auto-detect GPU instance type from node labels
+GPU_INSTANCE_TYPE=$(kubectl get nodes --request-timeout=30s \
+  -o jsonpath='{.items[*].metadata.labels.node\.kubernetes\.io/instance-type}' \
+  2>/dev/null | tr ' ' '\n' | grep -v "cpu" | sort | uniq -c | sort -rn | head -1 | awk '{print $2}' || echo "")
```
dasbatta, line 123: Please avoid silently reducing restore coverage to the most common instance type. With two H200 nodes and one B300 node, this pipeline selects only the H200 type, so the B300 node never attempts restore, while the final summary says all GPU nodes were validated. Collect all intended Ready/schedulable GPU nodes, or make a single-group selection explicit and scope the summary to that group. GPU_NODE_LABEL is documented above but currently unused, so it cannot select the omitted group either.

k8s-training/filesystem-csi-validation/05-run-checkpoint-restore-test.sh
```diff
@@ -0,0 +1,483 @@
...
+        cpu: "2"
+  volumes:
+  - name: checkpoint-storage
+    persistentVolumeClaim:
+      claimName: ${PVC_NAME}
+EOF
+
+    wait_for_pod "${READER_POD}" 600 || true
+
+    CROSS_LOGS=$(kubectl logs --request-timeout=30s -n "${TEST_NAMESPACE}" "${READER_POD}" 2>/dev/null)
```
dasbatta, line 436: Non-blocking: please guard this log capture and collect best-effort pod status/events when it fails. With set -e, a failed kubectl logs exits here before the diagnostic branch, suppresses the retrieval error, and triggers cleanup. That makes a pod which never starts, for example due to a mount or image-pull failure, difficult to diagnose. The same pattern appears in the writer and same-node reader.

dasbatta (changes requested):
Adam, the checkpoint save/load and checksum comparison look sound. Reduced-size local CPU tests correctly rejected modified, truncated, and missing checkpoints.

Requesting a small rework to node selection before merge. The script currently checks only the most common GPU instance type, then reports coverage of every GPU node. Please either include all intended GPU nodes or make the selected scope explicit in both configuration and the result. The documented GPU_NODE_LABEL override also needs to be implemented or removed.

I also left a non-blocking suggestion to preserve pod diagnostics when log retrieval fails. The pinned container, default 4 GiB payload, and actual cross-node CSI execution still need runtime validation.

PR #1146, opened 2026-08-24
Add Kueue support to k8s-training

aaroniscode:

Adds Kueue (Kubernetes-native Job Queueing) support to the `k8s-training` solution.

Kueue’s admission controller holds queued Jobs until their configured quota is available. When a Job requests topology-aware placement, Kueue also checks that capacity is available in the required Nebius GPU topology before allowing Kubernetes to schedule its Pods.

- Adds a reusable Kueue Terraform module using the pinned upstream Helm chart.
- Enables topology-aware scheduling using the documented Nebius GPU topology.
- Creates a Kueue `ResourceFlavor` for each GPU node group.
- Allows users to configure `ClusterQueue` and `LocalQueue` quota and access policy.
- Includes usage documentation and an editable GPU queue-policy example.
- Exercises Kueue during the existing Terraform test-mode apply.


k8s-training/variables.tf
```diff
@@ -404,6 +404,38 @@ variable "enable_kuberay_service" {
...
 }
 
+variable "kueue" {
+  description = "Optional Kueue installation with topology-aware scheduling enabled by default; queue policy remains explicit."
+  type = object({
+    enabled                   = optional(bool, false)
+    chart_version             = optional(string, "0.19.2")
+    namespace                 = optional(string, "kueue-system")
+    timeout_seconds           = optional(number, 300)
+    topology_aware_scheduling = optional(bool, true)
```
IronhandedLayman, line 414: Are we confident that TAS should be on by default? Cyril has some stuff already written for this to test in the ml-cookbook, I will use this to test against that.

k8s-training/variables.tf
```diff
@@ -404,6 +404,38 @@ variable "enable_kuberay_service" {
   default     = false
 }
 
+variable "kueue" {
+  description = "Optional Kueue installation with topology-aware scheduling enabled by default; queue policy remains explicit."
+  type = object({
+    enabled                   = optional(bool, false)
+    chart_version             = optional(string, "0.19.2")
+    namespace                 = optional(string, "kueue-system")
+    timeout_seconds           = optional(number, 300)
```
IronhandedLayman, line 413: 5 minutes is a common timeout default but we may be able to shorten it... I've seen Kueue come up in a minute or sooner. Did it come up slower for you?

modules/kueue/main.tf
```diff
@@ -0,0 +1,67 @@
...
+  yaml_body = yamlencode({
+    apiVersion = "kueue.x-k8s.io/v1beta2"
+    kind       = "ResourceFlavor"
+    metadata = {
+      name = each.key
+    }
+    spec = merge(
+      {
+        nodeLabels   = each.value.node_labels
+        topologyName = var.topology_name
```
IronhandedLayman, line 58: At Nebius, are we certain to only have a single topology, and every ResourceFlavor attaches to the same topology? It may sound like scope creep, but I am thinking about our Physical AI customers who may want separate topologies for RTX clusters and non-RTX clusters.

IronhandedLayman (approved):
I think we should have a flag for gang scheduling broken out, the ability to set different topologies for different resource flavors, and we should tests autoscaling with Ray if that is something we want to support. But as a first pass, this should be fine: while the first item will cause some minor headaches, the other two can be set using Helm values on the Kueue chart.

PR #1110, opened 2026-07-30
k8s-training: add production GB300 NVL72 rack support

saurabh22patel:
## Purpose

Add a production-oriented GB300 NVL72 deployment path to `k8s-training`.

The implementation models GB300 capacity by rack, adds safe rollout defaults and validation, and preserves the existing deployment path for other GPU platforms.

### Rack-aware GB300 infrastructure

- Added `gb300.rack_count` to configure the required number of racks.
- Create one fixed Managed Kubernetes node group and one NVLink instance group per rack.
- Model each rack as 18 nodes with four GPUs per node, for 72 GPUs per rack.
- Generate stable sequential rack keys such as `rack-001` and `rack-002`.
- Support an optional InfiniBand GPU cluster for one rack and require it for deployments with two or more racks.
- Disable the generic GPU node-group path when GB300 racks are enabled.
- Add outputs for node groups, NVLink instance groups, rack node and GPU counts, and fabric selection.

The GB300 node groups use:

- Platform: `gpu-gb300`
- Preset: `4gpu-112vcpu-800gb`
- OS: Ubuntu 24.04
- Driver preset: `cuda13.0`
- Default boot disk: 1 TiB

Local NVMe passthrough can be enabled for kubelet ephemeral storage.

### Rollout behavior

- Set the minimum provider version to `0.5.232`.
- Added the shared `node_group_strategy` interface to CPU, generic GPU, and GB300 node groups.
- Require a zero-surge rollout strategy for GB300.
- Recommend replacing one node at a time:

  ```hcl
  node_group_strategy = {
    max_unavailable = { count = 1 }
    max_surge       = { count = 0 }
  }
  ```

Zero surge prevents an update from requesting capacity beyond the physical 18-node rack.

### Validation and guardrails

GB300 plans are rejected when:

- `rack_count` is not a non-negative whole number.
- Two or more racks are configured without an InfiniBand fabric.
- GPU autoscaling, preemptible nodes, or public GPU addresses are enabled.
- The rollout strategy is missing or permits surge nodes.
- MIG, custom GPU drivers, or the existing x86 NUMA presets are enabled.
- The existing HGX-specific NCCL test or unvalidated KubeRay profiles are enabled.

Capacity models such as 17+1 or 16+2 remain scheduling and operational policies over a complete 18-node rack. They do not reduce the node-group or NVLink instance-group size.

### Documentation

- Added single-rack and multi-rack examples.
[...]

PR #1148, opened 2026-08-24
Fix k8s-training validation when node group strategy is unset

aaroniscode:

The `k8s-training` tests fail during Terraform validation when `node_group_strategy` uses its default `null` value.

[PR #1110](https://github.com/nebius/nebius-solutions-library/pull/1110) added validation that reads `max_surge` and `max_unavailable` behind a logical `||`. CI uses Terraform 1.11.4, which still evaluates those attribute accesses and raises a null-reference error before the tests can run.

This change makes those optional attribute reads null-safe with `try()`.

### DIFF
--- a/k8s-training/variables.tf
+++ b/k8s-training/variables.tf
@@ -126,8 +126,8 @@ variable "node_group_strategy" {
   validation {
     condition = var.node_group_strategy == null || alltrue([
       for setting in [
-        var.node_group_strategy.max_surge,
-        var.node_group_strategy.max_unavailable,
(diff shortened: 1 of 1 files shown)


IronhandedLayman (approved):
do we need allure version bump to 15 here?

aaroniscode:
> do we need allure version bump to 15 here?

It was bumped in #1147 , let me rebase and re-run the tests

aaroniscode:
Tests are failing to do other issues. Merging as it's approved.

PR #1154, opened 2026-09-01
SCHED-1920: Remove DCGM hpc_job job-mapping plumbing

theyoprst:
## Problem

nebius/soperator#2923 removes the `hpc_job` label integration from DCGM metrics. The terraform plumbing that fed it becomes dead: the `dcgm_job_map_dir` variable, the `hpcJobMapDir` helm value, and the worker `hpc-jobs-dir` hostPath mounts.

## Solution

- Remove `dcgm_job_map_dir`, `hpcJobMapDir`, and the `hpc-jobs-dir` hostPath mounts from the slurm module and its helm value templates
- Keep `dcgm_job_mapping_enabled` and redocument it: it still selects soperator's dcgm-exporter chart over the gpu-operator's stock one

Pairs with nebius/soperator#2923. Safe to land in either order: once the soperator side ships, the removed values are inert (extra Helm values are ignored and the hostPath volume was unused).

## Testing

`terraform fmt` and `terraform validate` pass.

## Release Notes

Breaking: the `dcgm_job_map_dir` variable is removed from the soperator slurm module. `dcgm_job_mapping_enabled` now only selects which dcgm-exporter chart is installed.

### DIFF
--- a/soperator/installations/example/variables.tf
+++ b/soperator/installations/example/variables.tf
@@ -1482,8 +1482,8 @@ variable "allow_o11y_region_migration" {
   default     = false
 }
 
-variable "dcgm_job_mapping_enabled" {
-  description = "Whether to enable HPC job mapping by installing a separate dcgm-exporter"
+variable "dcgm_exporter_enabled" {
+  description = "Whether to install soperator's dcgm-exporter chart. When false, the NVIDIA gpu-operator's stock dcgm-exporter is used instead."
   type        = bool
   default     = true
 }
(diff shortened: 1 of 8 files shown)


soperator/installations/example/variables.tf
```diff
@@ -1483,7 +1483,7 @@ variable "allow_o11y_region_migration" {
 }
 
 variable "dcgm_job_mapping_enabled" {
```
ali-sattari: why not rename it to `dcgm_exporter_enabled`?

ali-sattari (approved):

asteny (approved):

PR #1172, opened 2026-09-09
SCHED-1905: Don't use NFS on L+ clusters and support WEKA

dstaroff:
## Release Notes

WEKA is supported for Jail and Jail sub-mounts:
- Added a module for pre-creating WEKA filesystems
- Supported `NETWORK_SSD` and `WEKA` backends for the Jail and sub-mounts
- Validated that only one WEKA filesystem is configured and that all participating nodes use WEKA-compatible presets
- Renamed `filestore_jail` and `filestore_jail_submounts` to `filesystem_jail` and `filesystem_jail_submounts`
- Validated Jail submount paths to prevent duplicate mounts and `/home` conflicts

Changes for NFS:
- Prevented NFS from being used on `L` and `XL` clusters, where it can become a performance bottleneck
- Allowed NFS for `/home` alongside WEKA on supported cluster sizes
- Moved enabled NFS configuration under the `spec` field

Both:
- Added Terraform tests covering NFS sizing restrictions, WEKA requirements, and mount-path validation


dstaroff:
https://github.com/nebius/soperator/pull/2970

ali-sattari (approved):

soperator/installations/example/variables.tf
```diff
@@ -227,17 +246,135 @@ variable "filestore_jail_submounts" {
...
+
+resource "terraform_data" "check_resource_presets_for_weka" {
+  count = local.weka_is_used ? 1 : 0
+
+  lifecycle {
+    precondition {
+      condition = alltrue(concat(
+        [
+          local.resources.system.sufficient["weka"],
+          local.resources.controller.sufficient["weka"],
```
theyoprst, line 367: Looks like this check will always fail when with weka is turned on, because controller will never be sufficient for weka: controller preset is 16vcpu in each sizing tier, but weka is turned off on this preset. Correct?

theyoprst: So all user will have is error "All nodes should have sufficient preset if WEKA is requested.", and no clue which ones offend. So the follow-ups:
- Can we make the error more descriptive?
- maybe we can also update sizing_tier, so when weka is turned on, it will bump preset up to 32vcpu automatically, so users will not need to do it manually in terraform.tfvars? (it will add a little complexity there though)

rdjjke: Maybe we can update the controller preset in big sizing tiers and add something like "Use sizing tier N or use >=32vcpu presets for all nodes"

soperator/modules/slurm/variables.tf
```diff
@@ -380,6 +380,7 @@ variable "filestores" {
     jail = object({
       size_gibibytes = number
       device         = string
+      backend        = string
```
theyoprst: backend is never used, is it OK?

soperator/installations/example/tests/jail_submount_paths.tftest.hcl
theyoprst, line 1: All these tests (in this folder) are not running in CI. Let's add them?

soperator/Makefile
```diff
@@ -80,6 +80,18 @@ terraform-test-modules:
...
 	done
 
+.PHONY: terraform-test-installations
+terraform-test-installations:
+	@echo 'Running Terraform installation module tests...'
+	@for installation_dir in installations/*/; do \
+		if [ -d "$$installation_dir/tests" ]; then \
+			echo "Testing module: $$installation_dir"; \
+			pushd "$$installation_dir" > /dev/null; \
+			terraform test -var-file ./tests/vars.hcl; \
```
theyoprst: there should be terraform init call before, otherwise it will fail

soperator/Makefile
```diff
@@ -80,6 +80,18 @@ terraform-test-modules:
...
+			popd > /dev/null; \
```
theyoprst: This loop (and the original loop above) never fail on Mac because:

```
SHELL = /usr/bin/env bash -o pipefail
.SHELLFLAGS = -ec
```

My make version is 3.81, and `.SHELLFLAGS` was added in 3.82. That's why `terraform test` return code is ignored, popd return code is 0, hence the whole check return code is 0.

Maybe add ` -e` to SHELL?

theyoprst:
This LLM finding can be copied as is :-)

> 4. README wasn't updated, and one stale section is destructive. README.md:255-277 is the "retain filesystems on destroy" runbook:
> 
> terraform state rm 'module.filestore.nebius_compute_v1_filesystem.jail[0]'
> 
> Those addresses no longer exist. A user following it gets Invalid target address, and if they push past it, terraform destroy deletes filesystems they meant to keep. Also stale tfvars examples at README.md:106,113,224,232,297,303 (none with the now-required spec.type) and test/README.md:64,91.
>

soperator/installations/example/variables.tf
dstaroff: 👀

soperator/modules/slurm/variables.tf
dstaroff: No, it's a leftover. I'll remove it

soperator/installations/example/tests/jail_submount_paths.tftest.hcl
dstaroff: ＋

soperator/Makefile
dstaroff: It may break the existing installation's state. That's why I didn't add CI run for this check.
Maybe it's worth:
1. Copying the `$installation_dir` to `${installation_dir}_test`
2. Performing the test
3. Removing `${installation_dir}_test`
Then it may be included to CI

soperator/Makefile
dstaroff: I personally wanted to use Just instead of Make here. Let's see the `SHELL` fix

soperator/installations/example/weka.tf
```diff
@@ -0,0 +1,15 @@
+module "weka" {
+  count = 0
```
theyoprst, line 2: That's weird. I don't understand how to use it; the README doesn't tell me. Do we readlly need this whole module?

soperator/modules/available_resources/preset.tf
```diff
@@ -27,7 +27,9 @@ locals {
 
   presets_cpu = {
     c-2vcpu-8gb = {
+      cpu_cores_raw          = 2
```
theyoprst: My llm couldn't find any usages:
> Dead preset fields. cpu_cores_raw / memory_gibibytes_raw added to all 21 presets (42 lines), referenced nowhere — I grepped .tf/.hcl/.tfvars/.md, zero hits. Presumably the input to a numeric rule that got hardcoded as 21 hand-written booleans instead; sufficient.weka is exactly cpu_cores_raw >= 32 for every row. New presets that forget the weka key throw This object does not have an attribute named "weka".

soperator/Makefile
theyoprst: You are probably not talking about CI (because CI is running on github runners and there are no existing installations there), but rather about running these tests locally. My LLM says `.terraform` files might be changed and it might require to run terraform init again to fix it. It suggests this:

```
TF_DATA_DIR=.terraform-test terraform init -upgrade -backend=false; \
TF_DATA_DIR=.terraform-test terraform test -var-file ./tests/vars.hcl; \
```
 
+ gitignore folder .terraform-test

dstaroff: > You are probably talking about running these tests locally

Yes, I usually run these test against some existing installation or somehow initiated example. So it should be done to support testing on both clean installation during CI, and locally on existing installations.

The solution with `TF_DATA_DIR` works.

dstaroff: > gitignore folder .terraform-test

`.terraform*` is already ignored on `soperator` module level

soperator/installations/example/variables.tf
dstaroff: [Changed](
https://github.com/nebius/nebius-solutions-library/pull/1172/changes/627dda95465632e66c69b92b91000ac3261f3028) to
```
L  = "32vcpu-128gb"
XL = "32vcpu-128gb"
```
for controller and 
```
M  = "16vcpu-64gb"
L  = "32vcpu-128gb"
```
for accounting.

soperator/modules/slurm/variables.tf
dstaroff: [Done](https://github.com/nebius/nebius-solutions-library/pull/1172/changes/b1c2220a4df2630e63ff324b74cbc1e90f50794c)

soperator/installations/example/tests/jail_submount_paths.tftest.hcl
dstaroff: [Done](https://github.com/nebius/nebius-solutions-library/pull/1172/changes/db52dc54fdbf9d19a910ffc346f66c7158a637ef)

soperator/Makefile
dstaroff: [Done](https://github.com/nebius/nebius-solutions-library/pull/1172/changes/834a3a6bd7b836e474cdc401b9b889cd554acb6e)

soperator/installations/example/weka.tf
dstaroff: TLDR: we need it.

[Wrote](https://github.com/nebius/nebius-solutions-library/pull/1172/changes/9fd67d7971a86084a91a537192880482f77e9465) the usage to the Soperator module README, covering its use-case and flow.

Also, fixed renamed `filestore -> filesystem` mentions in this README.

Also, cleaned up the formatting of our READMEs and [added](https://github.com/nebius/nebius-solutions-library/pull/1172/changes/171b961b1e2dc88f28a500b9fe9b1d1868ea0794) CI lint check as well as Makefile target to check it locally.

soperator/modules/available_resources/preset.tf
dstaroff: Yeah, also leftovers from the first attempt on deep region-dependent sufficiency check. [Removed](https://github.com/nebius/nebius-solutions-library/pull/1172/changes/a791eb92469f1a449c0f6a1b3abd1501c7d8ec8b) it.

soperator/Makefile
dstaroff: [Done](https://github.com/nebius/nebius-solutions-library/pull/1172/changes/834a3a6bd7b836e474cdc401b9b889cd554acb6e)

soperator/modules/slurm/terraform.tf
```diff
@@ -17,3 +17,7 @@ terraform {
 module "labels" {
   source = "../labels"
 }
+
+module "resources" {
```
theyoprst: My LLM says it is never referenced, dead code, remove then?

soperator/installations/example/weka.tf
```diff
@@ -0,0 +1,15 @@
...
+
+  source = "../../modules/weka"
+
+  iam_project_id   = data.nebius_iam_v1_project.this.id
+  k8s_cluster_name = local.k8s_cluster_name
+
+  fs = [{
+    name                 = "jail"
+    size_gibibytes       = 2048
+    block_size_kibibytes = 32
```
theyoprst: > installations/example/weka.tf:12 — block_size_kibibytes = 32 is silently discarded. modules/weka/variables.tf:15 has no such attribute; Terraform drops extra object attributes. The one worked example the README points at contains a dead knob — made a no-op in d116f987 without cleaning the example.

soperator/installations/example/variables.tf
```diff
@@ -338,80 +462,104 @@ resource "terraform_data" "check_nfs" {
...
-    version         = optional(string)
-    use_stable_repo = optional(bool, true)
-    size_gibibytes  = optional(number)
-    disk_type       = optional(string)
-    filesystem_type = optional(string)
-    threads         = optional(number)
+    enabled = bool
+    spec = optional(object({
+      version         = string
+      use_stable_repo = bool
```
theyoprst, line 494: > variables.tf:494 — use_stable_repo promoted from optional to required, but never plumbed. main.tf:628-634 omits it, so modules/slurm always uses its own default. Users must now supply a value guaranteed to be ignored.

soperator/installations/example/variables.tf
```diff
@@ -227,17 +246,138 @@ variable "filestore_jail_submounts" {
...
+        # No guardrail
+        ? true
+
+        # Shared submounts should not be mounted to /home if there's already NFS for that
+        : alltrue([
+          for sm in var.filesystem_jail_submounts :
+          (sm.mount_path != "/home")
+        ])
+      )
+      error_message = <<EOT
```
theyoprst: > variables.tf:306 and :375 use <<EOT, not <<-EOT — both new error messages render with 8 leading spaces. Sibling heredocs at :689,:846,:861 get it right.

soperator/installations/example/main.tf
```diff
@@ -228,7 +228,9 @@ resource "terraform_data" "check_variables" {
     terraform_data.check_slurm_worker_cpu_platform,
     terraform_data.check_nfs,
     terraform_data.check_nfs_exclusivity,
+    terraform_data.check_nfs_sustainability,
     terraform_data.check_jail_submount_paths,
+    terraform_data.check_weka_count,
     terraform_data.check_local_nvme,
   ]
```
theyoprst, line 236: > check_resource_presets_for_weka is missing from the check_variables gate (main.tf:224-234 lists the other two). Only reachable via module.k8s's depends_on, so it fires later than its siblings. Looks accidental.

soperator/test/README.md
theyoprst: > still stale
> 
> - test/README.md:68 — filestore_jail_submounts = [{ (removed variable) and its spec at :71-75 has no type. Fails twice over.
> - test/README.md:95 — filestore_jail_submounts again; :79,99 prose "the same filestore".
> - README.md:183-189 — the commented filesystem_jail example is missing the now-required type. That's your flagged line 113.

soperator/modules/slurm/terraform.tf
dstaroff: [Done](https://github.com/nebius/nebius-solutions-library/pull/1172/changes/dd10daafd0df1c0be847c7db0e547ec1054241eb)

soperator/installations/example/weka.tf
dstaroff: [Done](https://github.com/nebius/nebius-solutions-library/pull/1172/changes/dd10daafd0df1c0be847c7db0e547ec1054241eb)

soperator/installations/example/variables.tf
dstaroff: [Done](https://github.com/nebius/nebius-solutions-library/pull/1172/changes/124b5cba3b4c5debcb373c0fac9bae6802ba7766)

soperator/installations/example/variables.tf
dstaroff: [Done](https://github.com/nebius/nebius-solutions-library/pull/1172/changes/6f7d464afb14fc810295a7a80852dd83bab74226)

soperator/installations/example/main.tf
dstaroff: [Done](https://github.com/nebius/nebius-solutions-library/pull/1172/changes/b8fcfc98659a423febee6dc9c27814fd58f0e1b0)

soperator/test/README.md
dstaroff: [Done](https://github.com/nebius/nebius-solutions-library/pull/1172/changes/ef6404e6726b8e84b6855daa7770da719fc2e860)

Plus fixed missing `.spec.type` field in references.

dstaroff:
https://github.com/nebius/soperator/actions/runs/35226646732

dstaroff:
> https://github.com/nebius/soperator/actions/runs/35226646732

E2E failed on unrelated Docker scenarios:
- Docker container lifecycle uses local storage
- Docker containers can access GPUs

theyoprst (approved):

PR #1176, opened 2026-09-11
Use IMDS token endpoint in bastion cloud-init

ashestimer:
## Summary

- replace the deprecated /mnt/cloud-metadata/token file with the HTTP IMDS token endpoint
- let the Nebius CLI retrieve refreshed service-account tokens from IMDS

Use the HTTP IMDS token endpoint when configuring the Nebius CLI on bastion hosts.

### DIFF
--- a/modules/cloud-init/bastion-cloud-init.tftpl
+++ b/modules/cloud-init/bastion-cloud-init.tftpl
@@ -88,7 +88,7 @@ runcmd:
   - su - ${ssh_user_name} -c 'source "/home/${ssh_user_name}/.bashrc"' >> /tmp/csa-install.log 2>&1
   
   - echo "### Nebius CLI profile create" >> /tmp/csa-install.log 2>&1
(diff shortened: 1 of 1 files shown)


jadnov (approved):
Tested at 2ef4e44f on a live deployment (arch-sandbox / eu-north1). LGTM.

**Verified on the bastion host:**
- Profile is created with the new endpoint:
  `token-endpoint: http://metadata.nebius.internal/v1/iam/sa/token`
- `nebius iam whoami` authenticates successfully as `bastion-sa`
- `/tmp/csa-install.log` shows `profile "default" configured and activated`, no errors
- `cloud-init status --long` → `done`, `errors: []`
- `wg-quick@wg0`, `wgui_server`, `wgui.path` all active
- Apply: 13 resources added, cloud-init 3m38s, `test_mode` remote checks passed

**Note non blocking:**

   The deprecated `/mnt/cloud-metadata/token` file is still present on the
   image, so the old config did not hard-fail in my test. The value of this
   change is the refreshed token rather than a fix for an outright breakage —
   might be worth reflecting in the release note.

